PR #10, opened 2016-07-22
Added "Save last coordination function"

ReddoKiddo:
Will auto save the last player coordination in Coords.txt file


NecronomiconCoding:
Thanks, added.

PR #14, opened 2016-07-22
Fixed window title Status information

Meeplez:
I was thinking about making maybe a function to collect console status and then post, Or else the console floods with the player info, this is just for title anyways, the information can still be recieved elsewhere

This is also ui stuff, i dont think we need to wait for http requests, that should be put elsewhere, as this is client side stuff. ie, buttons, label info.
also i think report status like this is useless, uless its like a level up, so like if 

`if(level>curlevel )
{level=level+1;  Msg(" YOU LEVELED UP! %s %s " Pkmonstatus.exp, Pkmonstatus.level); }`

this also should not result in a dissconnect due to http, as the report of levels is referred to some int, which would be null or zero if if couldn't get the info

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Logic.cs
+++ b/PokemonGo.RocketAPI.Logic/Logic.cs
@@ -31,7 +31,7 @@ namespace PokemonGo.RocketAPI.Logic
 
         public async Task Execute()
         {
-            Git.CheckVersion();
+            await Git.CheckVersion();
 
             Logger.Write($"Starting Execute on login server: {_clientSettings.AuthType}", LogLevel.Info);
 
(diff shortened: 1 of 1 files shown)


NecronomiconCoding:
updated the title , fix the conflicts and i will accept :)

Meeplez:
i wish i could see conflits, as i just updated the bas to ferox, and changed these. here ill postmy improvement
in postloginexecute under true
 DisplayPlayerLevelInTitle();

and replace the whole function of  Displayplayerlevelintitle with
private async Task DisplayPlayerLevelInTitle()
        {
            do
            {
                await Task.Delay(1000);
                var playerStats = await _inventory.GetPlayerStats();// await _inventory.GetPlayerStats();
                var playerStat = playerStats.FirstOrDefault();
                //  if (playerStat != null)
                // {
                var message = $"Player level {playerStat.Level:0} - ({(playerStat.Experience - playerStat.PrevLevelXp):0} / {(playerStat.NextLevelXp - playerStat.PrevLevelXp):0})";
                System.Console.Title = message;
                //Logger.Write(message);
                //   }
                // await Task.Delay(5000);
            } while (true);
        }

NecronomiconCoding:
you have differnt changes in your posts. try to get a fresh fork and start over :)

PR #16, opened 2016-07-22
	Catch while walking Part 2

Ehnoah:
Sorry I splitted it


NecronomiconCoding:
put them together and i will happily accept.

PR #28, opened 2016-07-22
Fix for coords.txt invalid coordinate error

ReddoKiddo:
this add checks if the coords/file has valid information.... Someone could do a better job (feel free to) but this will atleast work


NecronomiconCoding:
thanks, implemented

PR #33, opened 2016-07-22
Better logic for using berry

ReddoKiddo:
Will now use if capture probablity is lower than 0.35 and pokemon has high CP OR pokemon has the min IV percent


MonsterMMORPG:
makes sense

Sylencer90:
agreed.

PR #38, opened 2016-07-22
added username in title, display current latitude and longitude at startup

marksterxxx:
added username in title,
display current latitude and longitude at startup instead of default.


NecronomiconCoding:
added thanks

PR #46, opened 2016-07-22
Evolve pokemons after сatching

TheBringerOfPain:


PocketMobsters:
Are you sure about this though? Don't people use the current evolve option to evolve on startup after using a lucky egg?

navidemad:
:-1:

TheBringerOfPain:
Yes, you are right. Just i don't have any lacky eggs and need an option like this.

Sylencer90:
I mean it can be added by the creator for sure, but hopefully default as false.

solenrus:
I do this manually with every update, this really needs to be an option in the settings.

Shroomzie:
@solenrus i cant seem to find the code you have pushed to this project for this.

PR #56, opened 2016-07-22
Catch AggregateException so we don't die.

trcjr:
Catch AggregateException so we don't die.


NecronomiconCoding:
finally. such a small fix that i never saw. my bad. :heart:

PR #76, opened 2016-07-22
External Config Files for Pokemon to evolve and keep

marksterxxx:
Will load ConfigPokemonsToEvolve.txt and ConfigPokemonsToKeep.txt in Config Folder,
This will make it easier for others to configure.


roflinger:
good one!

eLyiN:
agree

PR #79, opened 2016-07-22
Fixed typo for geodude in proto

Daylend:
Noticed this and it was bothering me.


Daylend:
@NecronomiconCoding Somebody undid this commit again.

PR #87, opened 2016-07-22
Throw PtcOfflineException on maintenance

lauw:


lauw:
Maybe remove the data text check and have InternalServerError always throw it? This makes sure it doesn't crash on JSON parsing

PR #96, opened 2016-07-22
Update Settings.cs

Owinger:
I made it easier to read.


NecronomiconCoding:
Thanks but currently we have other plans to outsource the configs. You can make that if you like :smile:

PR #111, opened 2016-07-22
Updated console message to reflect xp required for next level.

Jaerin:
This should change it to reflect the actual Xp progress needed for next level.


RoachxD:
I did not saw your PR first..
Did it already, also the Title would've been kind of long.

PR #114, opened 2016-07-23
Include current location in title and output

OverlordQ:


OverlordQ:
Ugh, commits, let me rebase

PR #126, opened 2016-07-23
Removed Delay Between movement, Added delay between pokemon catches, CatchWhileWalking Fix

AdnanCukur:
Removed Delay Between movement. This delay was created as a delay between pokestop teleports, there is no need for this now since we have humanwalking.

Added Delay between pokemon catches, as setting, defaults to 5.

Removed unnecessary call to executecatchallnearbypokemon when a pokestop is reached. This already gets called by the humanwalking when we reach the pokestop.

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Logic.cs
+++ b/PokemonGo.RocketAPI.Logic/Logic.cs
@@ -298,7 +298,6 @@ namespace PokemonGo.RocketAPI.Logic
                     Logger.Write($"Encounter problem: {encounter.Status}");
                 if (pokemons.ElementAtOrDefault(pokemons.Count() - 1) != pokemon) // If pokemon is not last pokemon in list, create delay between catches, else keep moving.
                 {
-                    Logger.Write("There's more pokemon here",LogLevel.Info);
                     await Task.Delay(_clientSettings.DelayBetweenPokemonCatch);
                 }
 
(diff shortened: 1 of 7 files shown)


PokemonGo.RocketAPI.Logic/Logic.cs
```diff
@@ -341,7 +347,6 @@ public static int GetXPDiff(int level)
 
                 await Task.Delay(1000);
                 await RecycleItems();
-                await ExecuteCatchAllNearbyPokemons();
```
cvax: Why is this line being removed?

cvax:
Why are you removing ExecuteCatchAllNearbyPokemons() from Logic.cs?

AlginMaduro:
I think that way it will be searching/catching pokemon WHILE its walking toards the pokestop. In the current code its walking to pokestop first, then searching for pokemons near the pokestop. and traveling toads the pokestop its not doing anything besides walking.

cvax:
@AlginMaduro wasn't it already searching/catching while walking before this commit though? Thought we were already passing in the function to HumanLikeWalking().

AlginMaduro:
@cvax u are correct. we are passing the function through... but its being called only once. So basically its you "start human walking", its looks and catches pokemon in the "first step" and then it never searches again untill the next "excecution". and the next "excecution" of the func is  when arriving at the pokestop.

PR #139, opened 2016-07-23
Added PokemonInfo to API for BaseStats,Level,MaxCP,Perfection. Fixed proto for cpMultipliers(int->float)

5andr0:
The CalculatePokemonPerfection function now calculates the perfect MaxCP
percentage based on BaseStats. I also had to fix the protocol to change
the cpMulitplier values from int to float values

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Inventory.cs
+++ b/PokemonGo.RocketAPI.Logic/Inventory.cs
@@ -83,7 +83,7 @@ namespace PokemonGo.RocketAPI.Logic
         {
             var myPokemon = await GetPokemons();
             var pokemons = myPokemon.ToList();
-            return pokemons.OrderByDescending(Logic.CalculatePokemonPerfection).Take(limit);
+            return pokemons.OrderByDescending(PokemonInfo.CalculatePokemonPerfection).Take(limit);
         }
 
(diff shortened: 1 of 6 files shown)


AdnanCukur:
what was wrong with the way it calculated perfection earlier ?

UgaBugaBoo:
What's the difference between the two commits ?

5andr0:
First some background info: [TheSilphRoad Pokemon CP formula](https://www.reddit.com/r/TheSilphRoad/comments/4t7r4d/exact_pokemon_cp_formula/)
`Max CP = 0.1* ((BaseAtk + IV) * (BaseDef + IV)^0.5 * (BaseSta + IV)^0.5 * (CpM+ACpM)^2)`
As you can see the AttackIndividualValue has a greater impact on the CP than def/sta. The CP will also increase more with pokemons having high attack values. It'll be less for Pokemons whos CP is based on a high def base value. That's why we have to take the base stats into consideration. The previous function was just a little bit off, now its 100% accurate!

OverlordQ:
You'll need to compile the proto file to generate a new Payloads.cs, otherwise it'll have no effect.

tensei:
he changed the payload.cs if you would look at the changes 
and ofc it's tested and works

navidemad:
Your payload changes crashed my appm invalid cast

PR #147, opened 2016-07-23
Replaced hardcoded pokemon not to catch with text file list.

Daylend:
I'm not sure why it was hard coded not to catch pidgeys and rattatas, but now you can manage those settings from the "ConfigPokemonsNotToCatch.txt" file. Pidgey and rattata are no longer blocked.


marksterxxx:
This is better, except the generated text file puts mewto in the list by default... better remove that.

Daylend:
Yeah I wasn't the one who put that in. Either way, you can't catch a Mewtwo.

PR #148, opened 2016-07-23
Don't waste berries when there are no pokeballs left.

cvax:
Prevent berries from being used when there are no pokeballs left.
Improve logging for encounter fails.


NecronomiconCoding:
fix conflicts pls

PR #150, opened 2016-07-23
check if cords are out of bounds & add stardust to the console title

VanStorm:
added a check for the coordinates to ensure that the latitude is with -90 & 90 and the longitude within -180 & 180.
Also add stardust to the title.

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Logic.cs
+++ b/PokemonGo.RocketAPI.Logic/Logic.cs
@@ -79,7 +79,7 @@ namespace PokemonGo.RocketAPI.Logic
             {
                 var xpDifference = GetXPDiff(playerStat.Level);
                 var message =
-                    $"{playerName} | Level {playerStat.Level}: {playerStat.Experience - playerStat.PrevLevelXp - xpDifference}/{playerStat.NextLevelXp - playerStat.PrevLevelXp - xpDifference}XP";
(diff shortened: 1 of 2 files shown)


Spegeli:
The cord check is not correct because latlngis a string so we can't use >=

VanStorm:
thx for the info - gonna fix it.

PokemonGo.RocketAPI.Logic/Logic.cs
```diff
@@ -79,7 +79,7 @@ public static float CalculatePokemonPerfection(PokemonData poke)
             {
                 var xpDifference = GetXPDiff(playerStat.Level);
                 var message =
-                    $"{playerName} | Level {playerStat.Level}: {playerStat.Experience - playerStat.PrevLevelXp - xpDifference}/{playerStat.NextLevelXp - playerStat.PrevLevelXp - xpDifference}XP";
+                    $"{playerName} | Level {playerStat.Level}: {playerStat.Experience - playerStat.PrevLevelXp - xpDifference}/{playerStat.NextLevelXp - playerStat.PrevLevelXp - xpDifference}XP Stardust: {_playerProfile.Profile.Currency.ToArray()[1].Amount}";
```
marksterxxx: Stardust value will not update.

VanStorm: I see no reason why it shouldn't work. Since the profile is updated whenever the task is running. Otherwise the xp shouldn't work either.

NecronomiconCoding:
fix conflicts pls

PR #153, opened 2016-07-23
Cleaned up navigation logic. Cached inventory requests

AlginMaduro:
*Navigation logic is now using microsoft library for location.
*Now caching inventory for 30 seconds.(this was being called waaaay too often). 
GetBestBall() was fetching the whole inventory 4x in a row...


Spegeli:
Seems you missed something.

Typ- or Namespace GeoCoordinate not found ;-)

AlginMaduro:
added reference in PokemonGo.RocketAPI.Logic/PokemonGo.RocketAPI.Logic.csproj, so it should be there. Reload ur C# project.

ThomasSRL:
Nice fixes! Would test if I wasn't softbanned atm...

PR #165, opened 2016-07-23
.CSV Export all Current Pokemon

problemz:
This will write a file named "PokeList.csv" in the Direction where you started the bot.
It exports all Pokemon you own and write them as semicolon separated values in the File.
They are ordered by Perfection. You can add a Filter to change the order.

image


NecronomiconCoding:
fix conflicts

PR #183, opened 2016-07-23
Update FortLureInfo proto

OverlordQ:
Update 'known' second parameter in FortLureInfo.  Taken from https://github.com/AeonLucid/POGOProtos/blob/eeccbb121b126aa51fc4eebae8d2f23d013e1cb8/src/POGOProtos/Map/Fort/FortLureInfo.proto


NecronomiconCoding:
fix conflicts pls

PR #184, opened 2016-07-23
Update ConfigPokemonsToKeep.txt

deknarf:
* Added rare and ultra rare Pokemon
* In appropriate PokeDex order (first>last)
* Fixed typo: Alahkazam (should be Alakazam)

### DIFF
--- a/PokemonGo.RocketAPI.Console/Configs/ConfigPokemonsToKeep.txt
+++ b/PokemonGo.RocketAPI.Console/Configs/ConfigPokemonsToKeep.txt
@@ -1,7 +1,18 @@
-Dragonite
+Venusaur
 Charizard
-Zapdos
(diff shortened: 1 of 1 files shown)


solenrus:
this shouldn't work cause of differences in allenums, not sure

ThomasSRL:
Wasn't this just updated by @solenrus ?

solenrus:
also Blastiose is misspelled,
@thomasSRL unfortunately had to revert, cause I was too slow on all the pulls.

ThomasSRL:
@solenrus 
Are there any missing ones according to you? Perhaps you can create a new pull?

I'm still hoping for the radius patch to get re-opened. Don't want to start in central park to end up in Tokyo...

PR #196, opened 2016-07-23
Some Cleanup

DurtyFree:
Cleaned some code & variable naming


AdnanCukur:
I Approve, is it tested ?

DurtyFree:
Its tested, code logic is also untouched.
Will do some cleanup on the logic project too.

AdnanCukur:
Logic.cs is a master clusterfuck right now, we should split it up in several parts.

PR #206, opened 2016-07-23
Added random delay

Nyankoo:
Implemented random delay to catch, pokestop, evolve, recycle and transfer methods. Can be disbaled in Settings.
Default is on.


ThomasSRL:
Reminds me of you runescape scripts :)
Doubt it'll come down to timing of clicks when they use the banhammer though.

AdnanCukur:
This can get ugly pretty quick. We shouldnt have to type the From/To, you could create an extension for Task called RandomDelay and make it percentage based.

PR #239, opened 2016-07-23
Added few more humanizing features

AlginMaduro:
Features include:
- Randomizing throws, so not getting a perfect throw each time.
- Walking to the pokemon u are catching. (right now it catches a pokemon thats +/- 100 meters away)
- Added waits when choosing berries, choosing pokeballs 
- fixed issue where an eve with 0% chance of capture used all my balls
- fixed issue where trying to catch pokemon when catch status in error state

~~So Apparently my PR #153 was not merged by someone and merging to master branch removed the fixes... which means this PR also contains the changes to that PR~~

Merge request is derping a bit caluclating the diff

### DIFF
--- a/PokemonGo.RocketAPI.Console/App.config
+++ b/PokemonGo.RocketAPI.Console/App.config
@@ -55,7 +55,7 @@
       <setting name="TransferDuplicatePokemon" serializeAs="String">
         <value>True</value>
       </setting>
-      <setting name="DelayBetweenMove" serializeAs="String">
+      <setting name="DelayBetweenPokemonCatch" serializeAs="String">
         <value>5000</value>
       </setting>
       <setting name="UsePokemonToNotCatchFilter" serializeAs="String">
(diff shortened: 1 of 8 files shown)


AlginMaduro:
fml... gotta merge it... :'(
will do that a bit later

BioMuell:
Pls dont merge this, there is no need for that.
Currently there is no information about anti cheat and the like..
this would change nothing just slow the exp gain down....

Or add a option for this "feature" ....

emreycolakoglu:
+1 for option

Tyyrn:
It feels like this is a point where we need to decide where to go.
Should this be a bot that is close to human behaviours or maximize speed of farming/exp grinding?

You could argue there was no need for walking instead of teleporting (other than eggs, that are not yet implemented to be used) or adding any delays and it´s just slowing down the speed of getting exp if you "walk".

DaniilSokolyuk:
Good as option

Nyankoo:
Should be in there, with the option default on.

NecronomiconCoding:
conflicts

PR #240, opened 2016-07-23
Made keepPokemonsThatCanEvolve configurable

Gezan:


iamD2:
thats a nice one! , thanks for that man!

mczee:
Finally I don't have to rewrite both of those lines over and over again

CTDragon:
I did this last night, but I realized that the logic for Inventory.GetDuplicatePokemonToTransfer needs to be re-written first.
It needs to be written in such a way that the list is generated with the same code, then pokemon that can evolve are removed. That way we are sure the logic matches and if it is ever updated, we don't have to update it on both sides of that "if" statement. Keep it DRY.
Since it wasn't DRY, I wasn't sure if the logic was complete without tracing it. Inventory.GetDuplicatePokemonToTransfer should be checked before merging this pull request.

NecronomiconCoding:
fix conflicts pls

PR #253, opened 2016-07-23
Added option to KeepMoreItems and converted filter to simple Dictionary

CTDragon:
Added option to KeepMoreItems to keep larger quantities of common items when recycling
Reduced data structure of ItemRecycleFilter to a simple dictionary instead of an array of key value pairs


roflinger:
not that nice, i prefer the current

NecronomiconCoding:
fix conflicts pls

PR #261, opened 2016-07-23
Google Refresh Token Value Fixed

decryptionz:
Fixed token ID value for new user's for authenticating Google Accounts

### DIFF
--- a/PokemonGo.RocketAPI.Console/App.config
+++ b/PokemonGo.RocketAPI.Console/App.config
@@ -29,7 +29,7 @@
         <value>pw</value>
       </setting>
       <setting name="GoogleRefreshToken" serializeAs="String">
-        <value />
+        <value>SESSION</value>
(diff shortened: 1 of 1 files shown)


NecronomiconCoding:
what should that fix?

decryptionz:
The app.config value for google refresh token so people don't get kicked out of the authentication after 30 minutes

NecronomiconCoding:
You dont get kicked out. The Auth Token is saved in appdata. Are you sure its the issue?

decryptionz:
It's more for identification for new players? In ferox's build the value was identified in app.config for automatic login verification for google accounts

decryptionz:
since there is two acronyms for authentication in app.config google and ptc. I saw this was messed up so i fixed it so you can take your token and replace SESSION and automatically login then having to put a code into google.com/device

PR #274, opened 2016-07-23
Move google auth token to "GoogleAuth.txt"

cvax:
Instead of having to recreate tokens every time you move directories, move token save to GoogleAuth.txt.


NecronomiconCoding:
can you move it to the Configs folder and also use an .ini extension. thanks

cvax:
Done

NecronomiconCoding:
you forgot to remove the googleauth from settings and also the save function etc.

PR #294, opened 2016-07-23
Add Gained Xp in catch message

giesti:
How to solve merge conflicts,, if at the time of creating the pull request there are no conflicts?


cvax:
Why are you making a GoogleAuth.ini default? It doesn't need any. People generally aren't going to have an auth token to put into the bot before running the bot at least once to generate the GoogleAuth.ini

giesti:
For backwards compability. My auth token was saved in the settings. But the code does not look there anymore.
So I needed to go search in the code where that file should be located. Because everyone that already runned the program, has their token, they do not want to have a browser opened.

cvax:
Takes 2 secs to reauth with the browser the bot automatically opens and device code already copied to clipboard.

Takes far more time for a user to find the old %AppData%\Local\PokemonGo folder, find the right one for their directory, open it up, copy the token value, then finally come all the way back and paste it into an empty .ini.

Pakobbix:
but if they download the Zip with updates, and compile the bot, they now only need to copy and paste the .ini to the new Folder first step into copy/paste whole files instead of checking the app.conifg for changes, enter everything manually etc.

cvax:
User has no need to copy paste .ini anywhere. It is auto generated when it doesn't exist.

giesti:
It checks if file exists and not empty. So creating this default file does not break anything.
This is not the change for this pull!

PocketMobsters:
GoogleAuth.ini should not be changed

PR #305, opened 2016-07-23
[Feature] Add ItemList in config files

marksterxxx:
changed .txt to .ini
removed config folder from project since it is now autogenerated w/ values
config files no longer get replaced every build/upgrade


NecronomiconCoding:
fix conflicts

PR #312, opened 2016-07-23
Pokemon are evolved before being transferred, and transfer less often

Daylend:
Items/pokemon will only be evolved/transferred once every 5 stops.

Planning on adding a config option and a check to see if the bag is full at a later time.

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Logic.cs
+++ b/PokemonGo.RocketAPI.Logic/Logic.cs
@@ -335,6 +335,7 @@ namespace PokemonGo.RocketAPI.Logic
                 await Task.Delay(1000);
                 if(++stopsHit % 5 == 0) //TODO: OR item/pokemon bag is full
                 {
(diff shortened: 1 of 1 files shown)


solenrus:
good shit, I was using the 
```
if (_clientSettings.EvolveAllPokemonWithEnoughCandy)
                    await EvolveAllPokemonWithEnoughCandy(_clientSettings.PokemonsToEvolve);
``` pretty much everytime, but it's kinda individual

cvax:
Using evolve every 5 stops basically means you expend all your candies evolving something that might not be worth evolving if you haven't caught something good yet. Not sure it is desirable to waste candies like this.

Antidote78:
You could set how many pokestop you want to wait before evolving in the settings

Daylend:
@Antidote78 "Planning on adding a config option". So yeah, if this gets accepted, I will go further with this.

PR #315, opened 2016-07-23
Update Inventory.cs

qwikwit:


PocketMobsters:
You added a comment.. That I removed... I see we are at odds here.

TheCrimsonDevil:
For future reference, do you want us to comment our code or leave it blank? I'd prefer commented, but I could see it really getting out of hand with some people.

PocketMobsters:
If you'd like to comment your code that's fine, anything excessive will probably get removed. I'd prefer light comments over nothing, just a "hey this is why i did this" would be fine.

ilu01:
@PocketMobsters But isn't this a valuable and short comment as you expect it?

PR #317, opened 2016-07-24
Show next level up in actual hours and minutes

vingesus:
This code will make 0.75h to level up, go into 0.45h to levelup instead :)


PocketMobsters:
Will merge as soon as we overhaul the API (should be within the day)

xashenx:
If the unit is 'hour', then 0.75 is the correct value and changing it to 0.45 might be confusing.
If you want to display minutes, you'll have to get the sub-unit 'minute', so to have one of the following formats:
- "0h 45m"
or (if h == 0)
- "45m"

vingesus:
Thanks for the feedback @xashenx - Edit have been done now :)

PR #323, opened 2016-07-24
Various fixes + externally configurable items list

MichaelGatesDev:
- Changed 'optimize code' to true, not sure if this actually works but I always use it.
- Commented various parts of the code to improve readability
- Added items list config file (located in \\Configs\\ after building)
- Added constant path strings for configuration files because there is no need to constantly redefine them
- Various file path changes
- Renamed config files to fit more standard English


CTDragon:
Nice, I was just about to start pulling the items into a config file.

sliverfishes:
You shouldn't enable "optimize" in a debug build because it makes things harder to debug. If you want speed then just build a "Release" build which has "optimize" enabled and debug symbols disabled.

Beware though, some things aren't getting rebuilt when needed while set to "Release" so you need to build "Debug" first and then "Release" if something fails to build.

MichaelGatesDev:
Thanks for the info @sliverfishes. I wasn't aware of that.

NecronomiconCoding:
fix conflicts

PR #324, opened 2016-07-24
Keep more great balls and hyper potions etc using KeepMoreItems setting

CTDragon:
Added option to KeepMoreItems to keep larger quantities of common items (great balls, hyper potions mainly) when recycling
Reduced data structure of ItemRecycleFilter to a simple dictionary instead of an array of key value pairs


CTDragon:
Also keeps less Razz Berries by default. This was set to 200... Considering changing the more option on this to ~80, but left it as the large number for now.

CTDragon:
Actually, following what I've noticed others are doing, I'll probably just add these to a config file so each user can modify them to their own preferences and change them as they level up.

CTDragon:
In fact, someone already made a file config in another pull request. I am just going to cancel this and if that pull request gets rejected for some reason, I will write it that way myself. This isn't very useful as is.

PR #336, opened 2016-07-24
Fixed logging when using IV > CP.

JoshYuen:
Updated the logger to output "best" based on your setting. The logger was still printing best based on CP and if you have IV > CP set it will show incorrect selection for best.


JoshYuen:
Merged this with most recent commit  802f65e

PR #338, opened 2016-07-24
Update Logic.cs

orangesuave:
Added calculation for Perfect Pokemon and Max CP % in transfer logic.


PocketMobsters:
This has already been added.

PR #348, opened 2016-07-24
Orangesuave be the very best

orangesuave:
Includes logic in the Evolve event to ensure only the top 1% of your pokemon are evolved.

This allows you to put all pokemon names in the evolve list without fear of losing your best of each type.


PocketMobsters:
Conflicts must be resolved. Additionally this was not even made a toggleable option in your pull request, thus making it pointless since some users do not want only the top 1% of their pokemon evolving.

PR #355, opened 2016-07-24
Small fixes for PokemonsNotToCatch

pattythesleeper:
Made a very small fix on PokemonsNotToCatch. Now it wont by default catch 

Zubat
Pidgey
Rattata


TheCrimsonDevil:
But I want to catch all those - Easy level ups!

Sylencer90:
Yeah i want to catch pidgeys for the easy exp.

pattythesleeper:
This function could easily be turned off and list is easy to change

PR #364, opened 2016-07-24
Improved handling for some edge cases with the maximum radius feature

Daylend:
A few users were confused when the program would endlessly list their highest Pokemon. This should hopefully fix that.

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Logic.cs
+++ b/PokemonGo.RocketAPI.Logic/Logic.cs
@@ -42,7 +42,6 @@ namespace PokemonGo.RocketAPI.Logic
             do
             {//test
                 var probability = encounter?.CaptureProbability?.CaptureProbability_?.FirstOrDefault();
-                
 
                 var pokeball = await GetBestBall(encounter?.WildPokemon);
                 if (pokeball == MiscEnums.Item.ITEM_UNKNOWN)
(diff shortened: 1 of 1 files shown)


qaqsd123:
Still keeps listing it. :(

Daylend:
@qaqsd123 Does it at least give a warning saying there are no PokeStops nearby?

qaqsd123:
Didn't get that far

cvax:
This PR should not be accepted.

- `Coords.ini` is for easy configuration of GPS coordinates and overrides `DefaultLatitude/Longitude` in `UserSettings.settings`
- `Coords.ini`'s whole purpose is defeated if you are walking from `DefaultLatitude/Longitude` to `Coords.ini`.

The default values are only supposed to be used when Coords.ini GPS location provided by the user is invalid and broken. See Client.cs lines 50-60.

Daylend:
Coords.ini is used to store the last location the bot was at. If limiting the distance was based on Coords.ini, it wouldn't be limiting the distance at all. It would only be able to travel to PokeStops x meters away from its current position, which is NOT what was intended. If you use the first coords found in Coords.ini, you would slowly drift away from your original position as the bot is reset, eventually having to put new coordinates in to keep the bot in the same place.

This PR prevents edge cases where you somehow get away from the predefined boundaries. This usually happens when a user forgets to clear their Coords.ini file before moving to a new start location as set in UserSettings.settings.

The point is to allow the user to save their position regardless of where they roam and continue where they left off, but keep them within a predefined area that doesn't change.

cvax:
You are right. How about renaming Coords.ini to LastCoords.ini then to avoid confusion?

Daylend:
Sure. Someone just needs to put up a PR. Feel free to if you'd like.

cvax:
Also, if we are already going to walk back to the start, shouldn't we be spinning PokeStops along the way instead of just catching Pokemon? Could be a real long walk if someone was jumping across the country and didn't change both coords.

Daylend:
Someone shouldn't be using this to walk across the country. This is for the cases where the bot is a minimal distance away from the radius. Otherwise they should see the message generated and restart the bot with the proper coordinates.

ThomasSRL:
Dont worry @Daylend , it seems perfect as is! Hope it gets accepted!

PR #365, opened 2016-07-24
PokeStop max travel distance was not calculated off correct GPS coords

cvax:
Currently it checks for pokestops against DefaultLatitude/Longitude instead of Coords.ini.
Coords.ini setting is supposed to take precedence over DefaultLatitude/Longitude.


Spegeli:
It is correc that he use the Default values for calculator otherwise you will leave the MaxTravelDistanceInMeters because it use the current vals as referenz instead the "center" point you set with the defaul values.

cvax:
The issue is we have two sets of coords competing against one another. Serves no purpose to have a default value and be put in a current location that is somewhere else. You guys are trying to reconcile Coords.ini's starting location versus some random default lat/long when these settings should have been the same setting from the start.

I am working on another commit to just kill the setting from UserSettings.settings and leave one setting in Coords.ini.

Spegeli:
Coords.ini will be always the last coord he was so when you start the bot again it will start from there
so the current DefaultValues are more as Center Point from where he should start checking the MaxTravelDistance.

cvax:
You are right.

PR #377, opened 2016-07-24
Better values and descriptions of functions

decryptionz:
Described functions in app.config and set values in settings.cs for
console of what you would only need realistically


PocketMobsters:
Conflicts must be resolved, please fix these and make a new pull request.

PR #403, opened 2016-07-24
Implement UseItem and useLuckyEggsWhileEvolving

Valmere:
now with a bool in the config!


l3uddz:
Possible to write to logger when the egg hatches (e.g. Pikachu hatched) ? This going to be a very popular feature, good job Valmere.

PR #408, opened 2016-07-24
Better item settings

ReddoKiddo:


LasaFawkes:
Those values add up to 375 without counting incenses, lucky eggs, lures and incubators. I think the sum should be something along the lines of 330 just to have some spare room for those items, no?

ilu01:
> Those values add up to 375 without counting incenses, lucky eggs, lures and incubators. I think the sum should be something along the lines of 330 just to have some spare room for those items, no?

I am with you in this matter. It has been proposed before, that Pokeballs and Potions may have a maximum sum, and always only the best are stored. e.g. Maximum = 100, if possible 100 ultraballs, if not then 100-x ultraballs, x great balls ... etc.

I am going to try this later if noone beats me to it.

ReddoKiddo:
I have been running this settings last night all night, and it was still able to take more items from pokestop in the morning... So I expect everything to work fine

ilu01:
@ReddoKiddo Well then, try simple math: Pokeballs = 150, Potions = 150, Revives = 75. Those alone are up to 425. How do you think, this may fit in your bags?

ThomasSRL:
This will highly depend on location and speed. I dont think these values are good. Should be under 250/350 depending on level.

PR #411, opened 2016-07-24
Better keep

kbm15:
This keeps pokemon on a better algorithm than absolute values.
Documented on issues.


AlginMaduro:
why would u remove _clientSettings.KeepMinIVPercentage and replace it with a hard value of 85?

kbm15:
Because my KeepMinIVPercentage is set at 50 and it would waste a lot of berries.
It is meant to be replace by some function that decides if the iv of the target is worth of a berry.
Something such as wild.IV*wild.CP/best.CP >= best.IV
But I dont know how can I code that. So meanwhile this works for duplicates and gets the idea to use the formula on berries too, even not being coded yet.

PR #414, opened 2016-07-24
Added Unicode support

TheCrimsonDevil:
Remember to change to a font that supports your language, otherwise it'll still show as ???.


sp1ti:
Now I feel stupid. Forgot the ";" that's why it failed for me ;___;

Thank you!

PR #425, opened 2016-07-24
Readme update

justh1n10:


nrmodesto:
I agree with that, it's much more organized in the wiki.

PR #441, opened 2016-07-24
InventoryCapacity as UserSetting

ilu01:
Pokeballs, Potions .. now as a function of InventoryCapacity

Possible future addition: Form a Pokeball, a Potion and a Revive group and only store the XX best items of a kind.


VanStorm:
Do you mean that if you for example got an inventory size of 500 because you bought additional space, the bot starts to use this space better without throwing away all items?

orangesuave:
Good idea. Same would be helpful for pokemon capacity. Also to display both on PostLoginExecute.

ilu01:
> Do you mean that if you for example got an inventory size of 500 because you bought additional space, the bot starts to use this space better without throwing away all items?

Exactly
But now we have a config file. Well.

PR #442, opened 2016-07-24
[Feature] External Items to keep list

marksterxxx:
-changed logic to create not existing config folder at run time instead of post build
-not creates .ini config files with default values only if not exist


TheCrimsonDevil:
I love you so much for that item config. <3

PR #451, opened 2016-07-24
[Feature] External Items to keep list

marksterxxx:
-Create Config folder at run time instead of post build
-Now loads itemstokeep.ini, renamed .txt -> .ini
-config files created with default values if not exist
-fixed file name exception for logger


marksterxxx:
merginf first

PR #452, opened 2016-07-24
Added feature to reset coords based on distance and time last run

CTDragon:
Added feature to reset coords based on distance and time last run

Also fixed bug with log causing crash in master branch


CTDragon:
Merged with latest master and reviewed Logger merge

NecronomiconCoding:
conflicts :)

CTDragon:
Resolved again

NecronomiconCoding:
So what exactly has been added?

CTDragon:
When the program is first run, if the user could have traveled realistically back to the default location from the last known location, it deletes the coords.ini file so the logic will start from the default location.

CTDragon:
The code essentially looks and says if (distanceBetweenDefaultAndLastKnownLocation/timeCoordsLastUpdated) < 80kmph), start at the default location, otherwise continue from last known location and continue as normal.

CTDragon:
I have tested it with the latest merges and it works

CTDragon:
The point is, I (and someone who submitted an issue) like to start from the default location because there are several stops right there. If I haven't run it in a while, I want to start from that default point where there are several stops, not somewhere else.

CTDragon:
image

I do want it to venture out from there though and not set too small of a max travel distance because all that area has is pidgey and ratatta. lol.

solenrus:
seems like a great feature to me

OverlordQ:
This is the stupidest 'realistic' idea ever.

CTDragon:
It's not a realistic feature, it's a start over without deleting the coords file feature...

PR #453, opened 2016-07-24
Username show

fewufu:
Show char name or Account name depending if using google account or not.
I use this for multiple char logins PTC accounts use the username in title, google account users char name


NecronomiconCoding:
This is more user preference than a feature.

PR #457, opened 2016-07-24
Minor changes

l3uddz:
Removed lucky egg spam
Removed 30minute wait after using a lucky egg.
Added EGG logging level
Fixed a logger crash due to backslash in date, also it stores a log for
every single day instead of every single hour/minute too.


NecronomiconCoding:
conflics :haha

PR #458, opened 2016-07-24
Fix the hours/minutes not showing correctly

vingesus:


nrmodesto:
Confirm it's working now with that commit..

vingesus:
Should work, tested it a couple of times, you got any problems? - As far as i know it worked before as well, but then it apperently got changed...

nrmodesto:
It's working without any problem.

GamespikeYT:
is it just me or is the recycling thing also very slow?

PR #459, opened 2016-07-24
Lucky egg changes

l3uddz:
Removed lucky egg spam.
removed 30 minute wait after using a lucky egg.
Added EGG log level.
Changed Logger to use an hourly log file instead of one every minute.


Ehnoah:
Can you add a Boost EXP Option?

Use Lure (on CD)
Use Egg (on CD)

To maximaze EXP for Catching.

PR #464, opened 2016-07-24
Added Multiple Response Payload Handling

HelloKitty:
This covers cases where all Response payloads in the ResponseEnvelope are of the same type TResponsePayload. Doesn't cover cases where they are different types though.


dontmesswithfreitag:
still does a restart on each action -> pokestop/catch

HelloKitty:
What do you mean? This has nothing to do with a pokestop/catch.

This is for properly parsing Response messages that should/may contain multiple response Payloads.

HelloKitty:
Closing as I added a much larger pr at #477 .

PR #467, opened 2016-07-24
KeepMinDuplicatePokemon can now be set to zero

Nordiii:
Made it possible to transfer Pokemon when you have only one of this type left.
Fixes an error when you wanted to transfer your best Pokemon(last one).

(In case you only want Pokemons reaching your criterion)

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Inventory.cs
+++ b/PokemonGo.RocketAPI.Logic/Inventory.cs
@@ -115,7 +115,7 @@ namespace PokemonGo.RocketAPI.Logic
             {
                 return pokemonList
                     .GroupBy(p => p.PokemonId)
-                    .Where(x => x.Count() > 1)
+                    .Where(x => x.Count() > _client.Settings.KeepMinDuplicatePokemon)
                     .SelectMany(
                         p =>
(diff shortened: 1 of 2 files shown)


solenrus:
its neat, but at the same time u wont be able to see how many candies u have for specific pokemon untill you get one meeting your criterion, still might be useful to some

Nordiii:
You'r right didn't think about this one!
Could dump this info into a File....
Since probably no one would use this when he wants to run around later

Thanks!

wetshrimp:
nordii i was looking for a solution to this, thanks!

now the only question is how we can deal with candies. my proposal is that we can code something that keeps 1 of certain types that are constantly evolved (pidgey, ratata, caterpie, etc)

Nordiii:
You could add them to "ConfigPokemonsToKeep", when you use auto evolve it will discard them anyways after evolving!

wetshrimp:
good point! im sure that will work but some evolutions will break the CP barrier (over 1k for me) like raticate and zubats and then ill be quite upset with 200 of them!!! :)

yamashi:
Master is deprecated.

PR #469, opened 2016-07-24
Fix recycling

OverlordQ:
I know this should go to develop, but this is kinda a showstopper bug.

If the TryParse succeeds, it puts the value into itemAmount and returns true. If it returns true, it catches the If and so puts the itemAmount to 100. This can cause your inventory to get full and you can't recycle.


marksterxxx:
+1 for this didn't notice

l3uddz:
+1, fixes recyling of items and farmed items logger output.

kbm15:
+1 It works

Antidote78:
+1 Tested and Working !

PR #468, opened 2016-07-24
Update logic.cs to showcase amount of candies

doggiekuk:
Amount of candies shown in DisplayHighest function descending.

PR #472, opened 2016-07-24
Added ability to view candies on Catch

doggiekuk:
Due to the lack of updating candies after catches, this is still a bit glitchy.

Anyone can improve please.


roteak:
Tested and working.

Calculates correctly after Catch and Transfers.

http://i.imgur.com/yHLmQcM.png

JoshYuen:
I feel like the info should be on transfers not catch. Because it should increment on transfer and then you can see new total. I think the catch log is already long enough and we are basically making it longer.

Any others want ot throw in some input?

JoshYuen:
PR #468 has similar logic maybe there should be a collaboration to merge the two PRs into 1.

yamashi:
Master is deprecated.

PR #473, opened 2016-07-25
Use InvariantCulture to parse GPX values

Rejaku:
GPX files weren't working with regional settings not using . as decimal seperator.
Thus, I have added CultureInfo.InvariantCulture to all Convert.ToDouble calls related to GPX.

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Logic.cs
+++ b/PokemonGo.RocketAPI.Logic/Logic.cs
@@ -15,6 +15,7 @@ using PokemonGo.RocketAPI.Logic.Utils;
 // ReSharper disable CyclomaticComplexity
 // ReSharper disable FunctionNeverReturns
 using System.IO;
+using System.Globalization;
 
 #endregion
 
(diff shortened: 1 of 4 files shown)


redongreen:
GPX is not detecting Pokestops along the way to the waypoint.

Rejaku:
Pokestops are detected and used if they are in the radius of the route point. With GPX, the bot follows the path and doesn't orient itself by the stops.
This sounds like a feature request to me, and has nothing to do with the issue here.

sloanja:
@redongreen, the GPX file must have the pathing go directly over the pokestop to register.  The route taken is precise and will not vary to hit the pokestop if you do not program it in precisely.  If you don't know the exact pokestop gps coordinates, you'll need to determine that by looking at Ingress (https://www.ingress.com/intel) or determining exact GPS coordinates when you physically visit the pokestop or other methods not described here.

yamashi:
Master is deprecated.

PR #475, opened 2016-07-25
Fixed a logic error in GetDuplicatePokemonToTransfer & reduced size of GetXpDiff

5andr0:
With KeepPokemonsThatCanEvolve=true it also kept all pokemons that couldn't evolve. Added KeepPokemonsThatCanEvolve to Settings


yamashi:
Master is deprecated.

PR #476, opened 2016-07-25
Fix: Decrement cached berry inventory count when using a berry

mrnitrate:
When using a berry the cached inventory was not being updated with the new decremented value.


JoshYuen:
Should probably check to make sure the UseBerry function actually successfully used a berry before decrementing.

AlginMaduro:
@JoshYuen That is indeed cleaner, but its cached for 30 seconds. So if it was not used correctly, it would only be invalid for 30 sec. But good remark

PR #480, opened 2016-07-25
Display distance and bearing for points of GPX file

sloanja:
This updates the logic.cs file to calculate distance and bearing when using GPX files.

### DIFF
--- a/PokemonGo.RocketAPI.Logic/Logic.cs
+++ b/PokemonGo.RocketAPI.Logic/Logic.cs
@@ -280,9 +280,12 @@ namespace PokemonGo.RocketAPI.Logic
                         while (curTrkPt <= maxTrkPt)
                         {
                             var nextPoint = trackPoints.ElementAt(curTrkPt);
-                            if (
-                                LocationUtils.CalculateDistanceInMeters(_client.CurrentLat, _client.CurrentLng,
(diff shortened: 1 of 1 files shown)


KiriharaSenpai:
But on last version of bot when i activate gpx function it keeps reloading itself

sloanja:
@KiriharaSenpai , you probably have a bad GPXFile format.  This program requires a very specific format.

Here's a sample that works with my build:

<?xml version="1.0"?>
<gpx creator="WTracks" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.topografix.com/GPX/1/1" version="1.1" xsi:schemaLocation="http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd">
<trk><name>My Local</name><trkseg>
<trkpt lat="50.730643752607115" lon="4.93199533224106"></trkpt>
<trkpt lat="50.73063208518487" lon="4.93205165863037"></trkpt>
</trkseg></trk></gpx>

As the project states, you can get the list of waypoints from wtracks.  I redid the XML from WTracks to fit this exact format, and it works flawlessly with default build 1.0.4.4.  Once I had the NecronomiconCoding default build, i altered it with the few additional lines to add the distance and bearing to the next waypoint.

KiriharaSenpai:
Thanks for the answer sir im gonna give it a try

yamashi:
Master is deprecated.

Issue #494, opened 2016-07-26
[Suggestion]Add Family Candies to Transfer or Catch stats


kowzzz:
It is already done in Spegeli but I prefer Necros movement and options. Was wondering if this could be added in easily because it gives a good idea when to stock up on certain pokemon to fully utilize the lucky egg buff.

yamashi:
Don't really understand what you mean ?

kowzzz:
http://i.imgur.com/NAM3lz7.png
At the end you can see family candies.

yamashi:
So it's just a log change correct ?

kowzzz:
Yes I believe so. I can go look into the code to see if I can find where it pulls it.

yamashi:
It's fine, I understand the request, it will be done.

kowzzz:
Thanks mate.

PR #503, opened 2016-07-26
#494: Implemented Family Candies display when transferring and catchi…

kaeus:
Not sure where the initial CP/IV display is anymore, or I would add it there as well. It does not seem to display for me anymore, so I at least added it to transferring/capturing events.


dbfy:
Yea at this point I'm waiting a few days, I spend more time merging with my changes then actually coding.

Btw it's in PoGoUtils\PokemonInfo

PR #512, opened 2016-07-26
fix division by zero on keeppokemonthatcanevolve

tehaon:
Someone verify this wont break shit.. I merged down from head fork and it's trying to push that as part of this.. I assume that's safe but really all I care about is the division by zero fix lol.


tehaon:
looks like someone beat me to fix

PR #517, opened 2016-07-26
add MaxCandyToEvolve setting. Don't evolve anything that takes more t…

rayn21:
Don't evolve anything that takes more than X candy.  Default to 12 to stick to Pidget/Caterpie/Weedles. Everything else, just transfer and save candy for powering up.


5andr0:
Redundant to ConfigPokemonsToEvolve.ini
You should add a commit to suppress that config when your value is set. But also disable your feature if it is set to 0. Not sure if this here is necessary, my ini is quite small and a better option because i would also evolve 50 candy pokes that are useless like zubat.

`Zubat
Pidgey
Rattata
Weedle
Caterpie
Krabby
Paras
Meowth`

rayn21:
Good point.  The ini is a better option.i wasn't aware of this setting. I am going to revert my change.

PR #545, opened 2016-07-26
fixed division by zero in case pokemon has no evolutions

blackms:


yamashi:
This is already fixed.

PR #562, opened 2016-07-26
Slight logger improvements

JoshYuen:
I also saw on the TODO there will be a logger change so this is just a temporary fix until then.

Reduced text used for logging because it because unreadable unless the console window took up 70% of your screen width.
image


LasaFawkes:
I love you so much for this :heart_eyes:

LasaFawkes:
Oh, I just noticed "TRANSFERED" is misspelled, it should have two R -> TRANSFERRED :)

PR #581, opened 2016-07-26
Give json settings the touch

DurtyFree:


PoGo.NecroBot.CLI/Program.cs
```diff
@@ -50,7 +50,7 @@ private static void Main()
 
             machine.SetFailureState(new LoginState());
 
-            GlobalSettings settings = GlobalSettings.Load("\\config\\config.json");
+            GlobalSettings settings = GlobalSettings.Load("\\configs\\config.json");
```
yamashi: let's keep config, people are going to lose their temper if we keep changing the directory :p

Navyguy330:
Yeah not really sure what this one was supposed to do?

PR #583, opened 2016-07-26
Implementation of Task UseAllIncubators

Bogenschuetze:
First pull request, not sure if i did it right.


yamashi:
You didnt do it right, all these commits from us should not be in here.

PR #595, opened 2016-07-26
[Feature] Rename Pokemon above KeepIvPercent with Iv value in name

kaeus:
Should help a few of the requests people have made. Makes it easier than trying to find even in the display once that gets redone.


mich3lam:
@kaeus if possible to make it also favorite them would be amazing!

PR #606, opened 2016-07-26
use incense

nongnoobjung:


PoGo.NecroBot.Logic/Tasks/EvolvePokemonTask.cs
```diff
@@ -18,8 +18,10 @@ public static void Execute(Context ctx, StateMachine machine)
             if (ctx.LogicSettings.UseLuckyEggsWhileEvolving)
             {
                 UseLuckyEgg(ctx.Client, ctx.Inventory, machine);
+                UseIncense(ctx.Client, ctx.Inventory, machine);
```
yamashi: Why here ?

Ilses:
That isn't complete yet -- you also need a separate request to encounter and catch incense pokemon. I can throw a PR up with those.

Issue #615, opened 2016-07-26
[Bug] Not evolving when enough candies


Rawrlex:
Pidgey as example has 206 candies, he is transferring instead of evolving for XP.

Following options:
  "EvolveAllPokemonWithEnoughCandy": true,

  "PokemonsToEvolve": [
    "zubat",
    "pidgey",
	"gloom",
	"caterpie",
	"metapod",
	"spearow",
	"weedle",
    "rattata"
]

PR #624, opened 2016-07-26
fix evolve using GPX

Wil32:


Wil32:
should close #615

Rawrlex:
So I have to enable GPX? And creating a path? I want to use it just normally without GPX.

PR #642, opened 2016-07-26
Use async/await in state machine

patrikwlund:
I've basically done one thing in many places: Replace .Wait() and .Result with proper awaiting.

The only other thing I did was removing the StateMachine.RequestDelay() method, and used Task.Delay() instead.

Note that I messed up the upstream rebase a bit, so the commits are listed twice.


Artud:
Fix conflicts.

PR #667, opened 2016-07-26
Added MultiProfile via CommandLine parameter

5andr0:
`cli.exe <Profile>`
if no parameter is given, the standard paths are beeing used.
if Profile is set, it will use configs/logs from _CurrentDir\<Profile>\_
any path specific functions should use **GlobalSettings.ProfilePath** and
**GlobalSettings.ConfigPath**

### DIFF
--- a/PoGo.NecroBot.CLI/Settings.cs
+++ b/PoGo.NecroBot.CLI/Settings.cs
@@ -73,19 +73,15 @@ namespace PoGo.NecroBot.CLI
     public class GlobalSettings
     {
         public static GlobalSettings Default => new GlobalSettings();
-
-        private static string GetAuthPath(string path)
-        {
-            var fullPath = Directory.GetCurrentDirectory() + path;
-            string folder = Path.GetDirectoryName(fullPath);
-            folder += "\\auth.json";
-
-            return folder;
-        }
+        public static string ProfilePath;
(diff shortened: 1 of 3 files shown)


Navyguy330:
what did the old version do? / this edit do?

PoGo.NecroBot.CLI/Settings.cs
```diff
@@ -100,7 +100,6 @@ public static GlobalSettings Load(string path)
             {
                 settings = new GlobalSettings();
             }
-            Logger.Write($"nigger {fullPath} {ConfigPath} {ProfilePath}");
```
Navyguy330: Why does that say nigger lol

5andr0: That was a debug output. Sorry for the n word, not racist ;) I've been using this for ages to mark my debug outputs

Navyguy330: I was just curious what it was.  is it the actual N word, or like an
abbreviation or something?

sub7even:
ProfilePath is not working for auth.json
Only config.json is loaded from CurrentDir\<Profile>\Config
The function AuthSettings() still checks for auth.json from CurrentDir\Config (without Profile)

PR #672, opened 2016-07-26
Add functions for catching lured/incense pokemon; implement catching lured pokemon.

Ilses:
Does not account for filters yet (do not catch type-x pokemon, etc). Refactors CatchPokemonTask.cs, add CatchLuredPokemonTask.cs.

Logic for using incense/detecting incense pokemon/etc is for another PR.


UntilSunrise:
Did the same on https://github.com/NecronomiconCoding/NecroBot/pull/673 but for both features

Ilses:
Cool, I'll close this one.

PR #682, opened 2016-07-26
Add GPS location of Pokestop to log output

sloanja:
This adds the GPS location of pokestops found when *not* using GPX Path files.


nfettinger:
I would think printing the first couple of characters of the pokestop to be more informational than coordinates, but maybe that is just me.

sloanja:
I wanted to use this information to help build a GPX path file, which is why exact coordinates are beneficial to me.  I can link a whole chain of them in my area.

PR #684, opened 2016-07-26
Updates to Navigation.cs

nfettinger:
Added (debug) print statements. Only show up if your log level is set accordingly.
Added jitter to walking speed to make it seem less bot like.


nfettinger:
This may require a submodule update of Ferox to deal with GetLongRandom being static

NecronomiconCoding:
conflicts

PR #691, opened 2016-07-26
Make the paths compatible with Linux

danscava:
I'm trying to run on Mono and had problems because the paths are using the '\' char. 
So I changed them to Path.DirectorySeparatorChar.


arianon:
why not use `Path.Combine` when joining paths?

PR #723, opened 2016-07-26
Added support for translations, customizing messages to your desire

DurtyFree:
+ added settings entry to set the language code for the language file you would like to use


GDlike:
This is amazing, but can you add more things to translate, will be possible?

Consolelogger.cs
Program.cs
Settings.cs
The states, 
the tasks,

PR #726, opened 2016-07-26
Egg incubation

patrikwlund:
Automatic use of egg incubators before every farming run, with setting for enabling (default=on).

Currently not executed in-between pokestops, but that would be a good next step. Simply add up the distances between stops and when enough of a distance (1km+) has been traveled, check incubators.

### DIFF
--- a/PoGo.NecroBot.Logic/Tasks/UseIncubatorsTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/UseIncubatorsTask.cs
@@ -1,0 +1,57 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using PoGo.NecroBot.Logic.Logging;
+using PoGo.NecroBot.Logic.State;
+using POGOProtos.Inventory.Item;
+
+namespace PoGo.NecroBot.Logic.Tasks
+{
+    class UseIncubatorsTask
+    {
+        public static async Task Execute(Context ctx, StateMachine machine)
+        {
+            // Refresh inventory so that the player stats are fresh
+            await ctx.Inventory.RefreshCachedInventory();
+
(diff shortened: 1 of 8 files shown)


PoGo.NecroBot.Logic/Tasks/UseIncubatorsTask.cs
```diff
@@ -0,0 +1,57 @@
...
+                    // Unlimited incubators prefer short eggs, limited incubators prefer long eggs
+                    var egg = incubator.ItemId == ItemId.ItemIncubatorBasicUnlimited
+                        ? unusedEggs.FirstOrDefault()
+                        : unusedEggs.LastOrDefault();
+
+                    if (egg == null)
+                        continue;
+
+                    var response = await ctx.Client.Inventory.UseItemEggIncubator(incubator.Id, egg.Id);
+                    Logger.Write($"Putting egg in incubator ({response.EggIncubator.TargetKmWalked - kmWalked:0.00}km left)");
```
yamashi: No Logger calls here.
Use machine.Fire and create a new event for incubators

PoGo.NecroBot.Logic/Tasks/UseIncubatorsTask.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                    unusedEggs.Remove(egg);
+                    await Task.Delay(500);
+                }
+                else
+                {
+                    Logger.Write($"Incubator status update: {incubator.TargetKmWalked - kmWalked:0.00}km left");
```
yamashi: Same as above

patrikwlund:
Updated with an incubation event

PR #727, opened 2016-07-26
fix compile warning

Sekioh:
Fix 'unreachable code' from double returns.


Sekioh:
Bad pull left a few files out of date, was fixed by someone else it appears.

PR #749, opened 2016-07-27
Adjustments for MultiProfiles

5andr0:
Fixed an issue where config/auth.json wasn't loading from the profile subdir.
Logs are written into profile subdirs to avoid writing conflicts. It was working before, but somebody broke it (https://github.com/NecronomiconCoding/NecroBot/commit/e62da1d63025700f10b214f72d412e3cc7c675ad)
User login prompt will now only pop up when there is no auth.json at the default or profile path. Entered logins will be stored in the newly created auth.json

### DIFF
--- a/PoGo.NecroBot.CLI/ConsoleEventListener.cs
+++ b/PoGo.NecroBot.CLI/ConsoleEventListener.cs
@@ -105,7 +105,7 @@ namespace PoGo.NecroBot.CLI
                 : "";
 
             Logger.Write(ctx.Translations.GetTranslation(TranslationString.EventPokemonCapture, catchStatus, catchType, evt.Id,
-                evt.Level, evt.Cp, evt.MaxCp, evt.Perfection.ToString("0.00"), evt.Probability, evt.Distance,
+                evt.Level, evt.Cp, evt.MaxCp, evt.Perfection.ToString("0.00"), evt.Probability, evt.Distance.ToString("F2"),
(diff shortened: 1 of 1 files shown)


hakkotsu:
Issue was introduced after this [commit](https://github.com/NecronomiconCoding/NecroBot/commit/e62da1d63025700f10b214f72d412e3cc7c675ad) causing all logs being stored in one file and bot throwing `System.IO.IOException: The process cannot access the file`

Ilses:
Thanks! I almost missed seeing this PR

hakkotsu:
I've closed my own PR because i missed this one. First [commit](https://github.com/NecronomiconCoding/NecroBot/pull/749/commits/8ac49ca8b39cc0a3513f3433021247f071bdbe49) should work, not sure about second

Ilses:
also add at Program.cs:44
subPath = subPath.Replace("/", "\\");
and everything should be good for the PR

5andr0:
Hmm why the / \ replace? Path.DirectorySeparatorChar should be fine. Profile names shouldn't be with slashes, we'll trust the user about that.
What's wrong with the 2nd commit? Works for me here, otherwise we won't be able to load auth.config from the profile config path. What's your solution to that? We can't use getcurrentdir

Ilses:
@5andr0 I used the / \ replace to dump bots in a subdirectory. Effectively, you can run with:
$> ./xxx.exe bots/profile

PR #753, opened 2016-07-27
Feature/reverse geo lookup

AdamWeishaupt1776:
grabs the formatted address of location, for a more useful startup warning message.


AdamWeishaupt1776:
unwanted? sloppy? incorrect implementation? Just curious if I should try a different approach, or focus on my own fork.

PR #775, opened 2016-07-27
Compile fixes, libraries update/consolidation, filled out pokemon lists.

DeadElle:
• Libraries update and consolidation
• Unnecessary references and libs removed
• Assembly infos fixed/updated
• Properly filled out pokemon lists


DeadElle:
going to resolve conflicts

PR #776, opened 2016-07-27
add random delay and remove unneccesary using

lifangjie:
I just want to make it more like human being as a bot should be.


AdamWeishaupt1776:
I feel like this should be an option, not default behavior.

OverlordQ:
You should separate these commits into housekeeping and functionality/logic changes.

NecronomiconCoding:
conflicts also seperate them

Issue #632, opened 2016-07-26
Lucky Eggs Being Wasted


Maynaards:
You should make it so if using lucky eggs is set to true, it waits to do evolutions and does loads at once

nfettinger:
+1

NecronomiconCoding:
what do you mean by  does loads at once?

fwlemos:
@NecronomiconCoding He means: doing mutiple evolutions at once (Around 60 is the ideal cap), to maximize the use of the lucky egg time. If it's set to true, the bot should capture enough pokemons/candies to make a load of evolutions, then pop a lucky egg.

Navyguy330:
it may say using one, but it never does if one is active, you can see your count stays the same

NecronomiconCoding:
added to milestone 0.3

bignuts:
+1 for this, also i think that 60 pokemon might be not ideal since evolving takes 2/3 sec each, lucky egg lasts 30 min..
Also it might be better to prioritize evolve over transfer for those who don't want to mass evolve,  so if u have enought candy and the Pokémon do not meet CP and IV requirements and it's on the list pokemontoevolve, it evolve it have without egg.
For those who want to mass evolve wait that pokemon bag reach x/250 .. 250/250 would be best!!! It should keep and evolve Pokémon that do not meet CP and IV requirements and it's on the list pokemontoevolve.
If this feature is added maybe the setting keeppokemontoevolve might be changed to fit everything.. Sorry for my english.

@NecronomiconCoding

PR #791, opened 2016-07-27
Add configurable minimum number of evolvable pokemon for lucky egg usage

bjwood:
Ensures that a minimum number of pokemon are ready to be evolved before using a lucky egg. See #632


maxbrueckl:
But maybe we should collect pokemon to evolve up to the number configured, then use egg and evolve all at once.

bjwood:
@maxbrueckl That's what this request aims to achieve. If UseLuckyEggsWhileEvolving is set, the actual evolution process gets skipped until we have reached the minimum number defined. Once we hit that amount, it uses an egg and processes all the evolutions at once.

maxbrueckl:
Yes, of course. I missed the else branch. my fault, sry

PR #792, opened 2016-07-27
CP per Pkm

diegomfdez3:
I have a lot of Pinsir above 1200 but i need put MinCP 1200 for other pkm. Can create a bracket with CP per Pkm? Too could be with IV per pkm.


diegomfdez3:
Srry, i wanted tell an idea but didnt know how put the correct brackets

diegomfdez3:
My deepest apologies.

PR #813, opened 2016-07-27
Replaces Console with Windows Form (for more functionality later)

sk8rdude461:
While in it's current state it looks almost identical to the Console, the form allows for us to add buttons that can be used to reload configuration files or potentially change the coordinates.

It also adds scrolling to the program when it's compiled and ran stand-alone.


UntilSunrise:
CLI should stay CLI. If you want a GUI then create a new project in the solution.
Also Windows Froms are not Mono compatible and would break linux compatibility.

sk8rdude461:
> If you want a GUI then create a new project in the solution.
I'm still relatively new to C# and Visual Studio. But wouldn't adding a new project require me to mimic most, if not all, of what's already in the CLI? Or would I be able to reference it?

PR #850, opened 2016-07-27
Retry at same PokeStop up to 40 times, if it gives 0 experience (softban bypass)

TheBorna:
Simple solution to softban, if pokestop gives 0 experience, it will loop the looting up to 40 times, which will clear the softban in around 1 minute.


mich3lam:
After the 40 times you will be unbanned lol..
it's a trick to bypass the softban. go to a pokestop, open, wipe, close, open, swipe, close. arround~30 times and the softban ends

TheBorna:
@mich3lam, it opens same pokestop 40 times and softban is bypassed, that's what it does.

solenrus:
you should make an additional check or two, because sometimes it bugs and you get 0 xp even when you're not softbanned, spinning 40 times every time you would get 0 xp would end up in trouble in long run

l3uddz:
maybe if it happens 2 or 3 times in a row, then activate providing it has been turned on in config?

TheBorna:
@splenrus, @l3uddz
I have added so it checks if the pokestop can be looted, that should prevent the 0 experience issue you mentioned you had.

PR #864, opened 2016-07-27
Fix mono/Linux compatibility

hirschmann:
I fixed some problems which prevented the bot from running properly on mono/Linux.

Changes:

- replace `System.Device.Location.GeoCoordinate` with `GeoCoordinatePortable.GeoCoordinate` (nuget package), because `System.Device.Location` assembly ist not available on mono/Linux
- replace hard coded paths with calls to `Path.Combine()`
- change `Console.OutputEncoding` to UTF8


LoloLaurentiu:
is this still working on v0.3.2 ?

PR #874, opened 2016-07-27
Get Levelup Rewards

cawk:
This code works just fine 
if you want you can change the way it is displayed up to you?

### DIFF
--- a/PoGo.NecroBot.Logic/Inventory.cs
+++ b/PoGo.NecroBot.Logic/Inventory.cs
@@ -41,7 +41,18 @@ namespace PoGo.NecroBot.Logic
             if (pokemon != null)
                 inventory.InventoryDelta.InventoryItems.Remove(pokemon);
         }
+        public async Task<LevelUpRewardsResponse> GetLevelUpRewards()
+        {
+            var GetData = await _client.Player.GetPlayer();
+            
+           
 
+            var ClientLevel = await _client.Player.GetPlayerProfile(GetData.PlayerData.Username);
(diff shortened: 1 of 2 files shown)


lorenzowalz:
@cawk So this eliminates the current issue where the inventory gets overloaded after leveling too many times (bot doesn't pick up those rewards)?

cawk:
as far as i am aware when getting level rewards you will get all of the rewards even if your bag is full it will go over (this is just from my experience)

PandaWalz:
@cawk That's what I was trying to say - that right now we get them but it overloads and won't throw the rewards away.

Ilses:
@PandaWalz At least as of yesterday, I don't believe we even got them. Thus, they couldn't previously be thrown away.

cawk:
@Ilses that is correct yes you would manually have to go on to the app and claim the rewards and you would miss some if you level up too quickly 
this claims them for you so no need to do anything

Ilses:
Doesn't build for me with the current master.

yamashi:
Yep this won't build, fix the build issues and you are in!

kanfkasndklasdn:
I've patched it to build with master. ¿How can I help?

kanfkasndklasdn:
Problem is that it recycles almost all that it receives

bignuts:
that is normal beacause u have set max amount in settings, but this is useful for getting incubator, lucky eggs, lure ect without logging.

cawk:
i will fix this for the latest version now wont take me too long to do so closing this for now

PR #896, opened 2016-07-27
Speed Control

nyirsh:
Most of Task.Delay calls are now configurable via ILogicSettings.
I only skipped the one in the GoogleLogin and the Thread.Sleep for the GoogleLogin in the PoGo.NecroBot.CLI because I feel like there's no point for the user in tweaking those two values.


PandaWalz:
@nyirsh You've got conflicts

ShawnHardwick:

PR #897, opened 2016-07-27
Softban bypass - check if pokestop can be looted (small improvement)

TheBorna:
Few users reported that I should also add check, here is simple fix it will check if the pokestop is looted already or not, if it isn't looted it will try to loot it 40 times, if it's looted it will continue with program.

This should prevent checking the pokestop if it's already looted.

Personally I never received 0 exp from pokestop but, just so we are safe.


NecronomiconCoding:
check conflicts pls

AdamWeishaupt1776:
IMO softban bypass should also implement logic to not catch any pokemon (otherwise you waste pokeballs/berries/etc.)

TheBorna:
Someone added check for 5 times which is pretty much useless and caused a lot of errors to users, as this way we check if pokestop can be looted or not.

PR #898, opened 2016-07-27
Added translation to Logger by injecting Context + some new translations

panthernet:
Context is used in logging only if it has been set. It is set after it has been created in Program.cs


panthernet:
Also have to deal with this for better translations. Please advice when you'll have time to do so.

ShawnHardwick:

PR #907, opened 2016-07-27
Egg incubation checks during PokeStop farming, and logging for hatch-events

patrikwlund:
**Edit: See new comment below**

Still doesn't log hatch-events though, since `ctx.Client.Inventory.GetHatchedEgg();` doesn't seem to return anything. Perhaps because the inventory (which includes pokemon) has already been refreshed?

### DIFF
--- a/PoGo.NecroBot.Logic/Tasks/UseIncubatorsTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/UseIncubatorsTask.cs
@@ -1,7 +1,12 @@
 ﻿#region using directives
 
+using System;
+using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
+using Newtonsoft.Json;
 using PoGo.NecroBot.Logic.Event;
 using PoGo.NecroBot.Logic.State;
 using POGOProtos.Inventory.Item;
@@ -12,6 +17,8 @@ namespace PoGo.NecroBot.Logic.Tasks
 {
     internal class UseIncubatorsTask
     {
+        private static readonly string IncubatorUsagePath = Path.Combine(Directory.GetCurrentDirectory(), "Temp", "Incubators.json");
+
         public static async Task Execute(Context ctx, StateMachine machine)
         {
             // Refresh inventory so that the player stats are fresh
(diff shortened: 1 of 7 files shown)


patrikwlund:
Since this PR hasn't been merged yet, I added additional functionality to it.

We now have hatch events for eggs. However, they are not based not actual hatch events from the server, because those turn up empty (for whatever reason). Instead, the client now remembers what the incubators stored previously and checks if there's a non-egg pokemon with that ID now (=hatched).

The incubator state is stored in: /bin/Temp/Incubators.json and doesn't require any kind of manual setup.

Not a very good solution, but it works.

patrikwlund:
Since no one else has tested this yet:

I've had this running since I pushed it. Here's a screenshot of me hatching a zubat in-between pokestops. This covers both the frequency-commit and the hatching-commit. Would still be good if someone other than me tried it out though.

incubator

Ilses:
I suspect 
private static readonly string IncubatorUsagePath = Path.Combine(Directory.GetCurrentDirectory(), "Temp", "Incubators.json");
will break with multiple profiles.

PoGo.NecroBot.Logic/Tasks/UseIncubatorsTask.cs
```diff
@@ -12,6 +17,8 @@ namespace PoGo.NecroBot.Logic.Tasks
 {
     internal class UseIncubatorsTask
     {
+        private static readonly string IncubatorUsagePath = Path.Combine(Directory.GetCurrentDirectory(), "Temp", "Incubators.json");
```
Ilses: I suspect this will break with multiple profiles.
private static readonly string IncubatorUsagePath = Path.Combine(Directory.GetCurrentDirectory(), "Temp", "Incubators.json");

5andr0:
Good find @ Ilses
You can use it like that:
```
        private static string IncubatorUsagePath;

        public static async Task Execute(Context ctx, StateMachine machine)
        {
                IncubatorUsagePath = Path.Combine(ctx.LogicSettings.ProfilePath, "Temp", "Incubators.json");
                
                string folder = Path.GetDirectoryName(IncubatorUsagePath);
                if (!Directory.Exists(folder))
                {
                        Directory.CreateDirectory(folder);
                }
```

You can also use the already existing Logs directory instead of Temp. Otherwise you would have to check for its existence and create it

Ilses:
I think I'd rather see:
```c#
private static readonly string IncubatorUsageFile = settings.ConfigPath + Path.DirectorySeparatorChar + "incubators.json"
```

patrikwlund:
Since it's neither a config nor a log, I think "Temp" fits better. It's not something you should include in your config backups.

Ilses:
Fair enough. I'd still like the var renamed, but now I'm just picking nits ;)
I'll try it out pretty soon and let you know how it works.

patrikwlund:
Made a commit for ProfilePath now. Also changed variable name to "...FilePath". A good compromise :)

solenrus:
using this with no issues, merged myself with the latests commit, good work!

Pakobbix:
can confirm.. it's pretty awesome to really hatch them while the bot runs and knowing the pokemon you got in your egg..

5andr0:
Ready to merge, thanks for the profile path integration. Good work

PR #908, opened 2016-07-27
Immediately Transfer Specific Pokemon on Catch

rayn21dev:
Is this useless with the new TransferFilter logic?  If so, please close and I need to revert my changes in my fork.


rayn21dev:
Reverting, better feature available.

Issue #899, opened 2016-07-27
Looting failed, possible softban


EasyAsABC123:
With latest code I get
```
[09:54:37] Looting failed, possible softban. Retry: 1/40
[09:54:38] Looting failed, possible softban. Retry: 2/40
[09:54:38] Looting failed, possible softban. Retry: 3/40
[09:54:39] Looting failed, possible softban. Retry: 4/40
[09:54:40] Looting failed, possible softban. Retry: 5/40
```

With previous code (older version) I just reopened on the same account and it loots just fine.

PandaWalz:
I believe if it goes to an already looted Pokestop that may be causing this. Looking, take a look too.

Hardell:
http://i.imgur.com/N7jqkjC.png

Yes, it checks already looted stops.

ridoknight:
same here 
![adsiz](https://cloud.githubusercontent.com/assets/20602186/17178038/8d1cad8c-541c-11e6-9f80-e00a0caa1d97.png)

crixon:
yes it's a bug!

erfg12:
Same here. I'm just getting spammed with Looting Failed messages.

Kristoffer88:
You have to revert this commit to fix it.

https://github.com/NecronomiconCoding/NecroBot/commit/0983e8af8b483a6562444677e8b01fad9072a977

EasyAsABC123:
@Kristoffer88 that doesn't look like it will actually fix the bug, it would just change how a softban is detected...it would still be failing to loot

Kristoffer88:
@EasyAsABC123 It seems to be working for me, i am looting right now

EasyAsABC123:
@Kristoffer88 interesting, it looted for me for hours and now is just broken and no amount of time resolves the issue now.

erfg12:
Reverted. Still have the issue.

mich3lam:
The commit is good, and the idea of how it's working is good. the problem is - it detects pokestops u cant take as 'softban' . this happens when u loot the same place, for exmaple : opened the bot for 1 min, closed it, and reopned it again > u cant take that pokestop

KiriharaSenpai:
doesnt related with that commit sometimes it occurs some server problem in game like that when i go in game theres nothing around except pokestop gym i think at these states it logs Looting failed

iBananaa:
Same here https://gyazo.com/05ad13dd521596c03f3a0508f951f055

[...]

PR #917, opened 2016-07-27
Bug Fix: softban detection was incrementing TimesZeroXPawarded incorrectly 

mrnitrate:
Bug Fix: softban detection was incrementing TimesZeroXPawarded when Experience was Awarded , instead needs to be when NOT awarded experience.


EasyAsABC123:
Resolves #899

Hardell:
`if (TimesZeroXPawarded > 5)`
should be 
`if (TimesZeroXPawarded < 5)`
as well. Otherwise it tries to clear softban when the counter is < 5, which makes no sense.

TheBorna:
It shouldn't even have timeszeroxawarded, someone added it to my code which caused erros. 
It's pointless to have it in loop which checks the same pokestop...

TheBorna:
Softban bypass - check if pokestop can be looted (small improvement) #897

Hardell:
This is still completely broken without fixing the other line. You also need to reset TimesZeroXPawarded after getting XP or you will be stuck in one part of the if statement forever.

PR #920, opened 2016-07-27
Bug Fix: FortUsedEvent needs to be called when TimesZeroXPawarded <= 5 not greater than

mrnitrate:
Bug Fix: FortUsedEvent was only firing when TimesZeroXPawarded was greater than 5 instead should fire when less than or equal to 5


Hardell:
TimesZeroXPawarded also needs to be reset to 0 after getting XP, otherwise you'll be stuck in the else block forever.

PR #981, opened 2016-07-27
Add indonesia translation file

hendrawd:


NecronomiconCoding:
please now include into .csproj

hendrawd:
Is that enough? Actually i have only little experience with C#

panthernet:
Yes, that's enough. Please update lang file with newly added translations. Thank you,

hendrawd:
@panthernet lang file here is translation.id.json right? Can you give a hint where is the newly added translations path? Actually i translate it from translation.en.json, but i didn't find it on the project, but only from release

clicknetcafe:
newly added Indonesia translations

https://gist.github.com/clicknetcafe/b92393e590ae3157cf49242cb6a8653e

PR #982, opened 2016-07-27
Add Lucky Egg

cawk:
Lucky Egg is used just over 30 minutes (too make sure its defiantly finished) it is used constantly
will add option for it now just thought to add the option


cawk:
Will Close And Update With Complete

PR #983, opened 2016-07-27
Use Lucky Egg Constantly

cawk:
Have the option to change in the config file
will check just over 30 minutes and put a new lucky egg on if there is one available


ShawnHardwick:

Issue #975, opened 2016-07-27
 (ATTENTION) PokemonInventory is Full.Transferring pokemons... 


ivicanikolic034:
I always get this message, and it is stuck to 0 transferred pokemons. My config file is ok, i think:

[config.zip](https://github.com/NecronomiconCoding/NecroBot/files/386874/config.zip)

Issue #930, opened 2016-07-27
Transfer not working properly anymore


GamespikeYT:

rayn21dev:
Can you provide some details?

Lovebucket25:
I also have this issue.I checked my account and I have like 45 Magikarp, and the bot keeps saying cant catch Pokemon bag full

GamespikeYT:
I have my settings set that if i have a duplicate that has an IV of lower than 99% and lower than 1400cp should be transferred. But after an auto transfer from the program, there are still some weedle's etc

GamespikeYT:
Did you put keeppokemonthatcanevolve = false, because i did, it works fine in the 0.1.7 release but not in the latest version of the code

Lovebucket25:
its default set to that

W1nd0x:
I got the same problem..

Issue #742, opened 2016-07-27
PokemonDoTransferList


MarkHuangZ:
Is that possible to have a PokemonDoTransferList? I turned the KeepPokemonsThatCanEvolve and EvolveAllPokemonWithEnoughCandy to true. It evolves all Pidgeys and other pokemons in Evolvelist; however, the bot left all second stage pokemons in the inventory... So I am thinking whether players need those second stage exp pokemons in their inventory or just transfer them all. If it is possible to have a PokemonDoTransferList, players can add pokemons in the list,  and call TransferDuplicatedPokemonTask. It would be so great to have.

Forgive me my poor grammar...

kingchenc:
+1

ShawnHardwick:
I'm not positive, but I believe that NecroBot should transfer the second stage pokemon as duplicates as long as they are under the CP and IV values.

Issue #613, opened 2016-07-26
Suggestion (Pokemons not to catch): Catch and transfer instead of skipping.


Caydenz:
It would be a lot better to catch the pokemon and transfer it afterwards instead of just skipping it.

kanfkasndklasdn:
This is what it's doing when you don't set the variable to true. Setting the variable to true skip those pokemon in order to catch quicklyer other pokemon so you get more candies in those.

Caydenz:
It's not doing that. If the CP or IV is higher than the threshold, the pokemon won't be transferred.

agp89:
So what you mean is like a list "PokemonsToTransfer" which does not respect the CP or IV threshold but still respects "KeepMinDuplicatePokemon"?

Caydenz:
Yep, that's what I mean. Something like that will be fine as well, probably even better.

srounet:
Same thing here:

    Encounter problem: PokemonInventoryFull

There are pokemons that you know, you don't want to keep, but still want to catch for xp.
So having a File where you can specify the names of the pokemons you want to "throw" no matter what CP or IV it is would be nice.
It would be the exact opposite of the "Mew, Mewtwo", which obviously, you want to Keep no matter what :)

salvob41:
This would be awesome!

Wil32:
close #492 if this get completed. its basically the same thing

bignuts:
if u set transfer duplicate pokemon to true and set it to 0, it does exactly what u mean, keeps only the one with IV and CP that meet requirements threshold.

We cannot remove pokemonnottocatch filter as a lot of us are using gpx pathing and if u don't filter some pokemon u're out of pokeball really fast!

srounet:
That's true but then, if I have **35 * Pinsir of 2kCp and 95% IV** and still want to get rid of them how can I do without messing with my other really rare pokemons with high IV and CP?

There is a list of "MustKeepPokemon no matter what", there should be the opposite "TransferPokemon no matter what".
That way you can still capture them for XP, candies and throw them away just after.

nenko91:
Good idea

PR #1000, opened 2016-07-27
Add PokemonsToTransfer

srounet:
Add a setting to force pokemon transfer using a PokemonsToTransfer setting.
This feature permit to still catch all Pokemons wich obviously is good for Candies and XP.
Still you want to transfer Pinsir, Rattata, Pidgey even if they are above both CP and IV.

This feature through a PokemonsToTransfer list permits to catch pokemon and later on transfer them keeping the current StateMachine Logic.

related issues:
- https://github.com/NecronomiconCoding/NecroBot/issues/975
- https://github.com/NecronomiconCoding/NecroBot/issues/930
- https://github.com/NecronomiconCoding/NecroBot/issues/742
- https://github.com/NecronomiconCoding/NecroBot/issues/613

### DIFF
--- a/PoGo.NecroBot.CLI/Settings.cs
+++ b/PoGo.NecroBot.CLI/Settings.cs
@@ -221,6 +221,11 @@ namespace PoGo.NecroBot.CLI
             PokemonId.Doduo
         };
 
+        public List<PokemonId> PokemonsToTransfer = new List<PokemonId>
+        {
+            PokemonId.Pinsir
+        };
+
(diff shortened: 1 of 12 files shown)


srounet:
Just noticed the **PokemonsTransferFilter** and think this code is complementary.
What **PokemonsTransferFilter** does is given a set of constants transfer pokemons, ofc you can set 0 cp and 0 IV for every pokemon you want to transfer.
It's way easier to just put those name you want to keep the same way you would ignore or evolve them.

ShawnHardwick:

We also discussed that this functionality already exists with the PokemonsTransferFilter

PR #1021, opened 2016-07-27
Remove path separator char prepended to profile, as this breaks Path.Combine

Ilses:
Barely larger than the Eevee typo


5andr0:
Also figured this out. I'm on it to push a pr

PR #818, opened 2016-07-27
Refactor Execute() in CatchPokemonTasks.cs to fix 3 bugs with catching incense Pokemon.

Ilses:
This PR supersedes PR #778 where the bugs are described. I have confirmed this refactor works when catching regular, incense, and lured Pokemon.

PR #1026, opened 2016-07-27
Refactor CatchPokemonTasks.cs execute function & fix incense

Ilses:
Fixes 3 bugs with catching incense Pokemon described in PR #778.

PR #1031, opened 2016-07-27
Display XP for Catches

rayn21dev:
Also includes #818 which fixed bugs with Pokemon catching - I couldn't test without this change.


Ilses:
Oops, I'd already updated #818 to #1026 .

rayn21dev:
That's alright, I actually didn't mean to include your fix in my pr but I'm
new to git and didn't realize all my commit would update my existing pr.
Either way, your fix and my change should not cause conflict.

PR #1049, opened 2016-07-27
Removed illegal characters from CLI csproj

DrDelay:
They look like fragments copied from a diff and caused an error opening the project in my IDE (not VS).


DrDelay:
OK you were faster.
Fixed in #4bbc330e79da4d697254c09e0e270529f5d9e9a9

PR #1051, opened 2016-07-27
fixed: Commata for "Incubated egg has hatched"

Pakobbix:


Pakobbix:
Got fixed, Panthernet was faster ^^

Pakobbix:
Got fixed, Panthernet was faster ^^

PR #1058, opened 2016-07-27
Evolvechange

skupfer:
Changed the settings from evolveABOVE to evolveBELOW-Iv-Value...

Is EvolveAboveIvValue really needed? You have EvolveAll for that. But EvolveBelowIv is more important maybe.

Or do we need both?


skupfer:
going to bed. till tomorrow I would ahve to merge alot more. nope ;)

PR #1067, opened 2016-07-27
Use lucky egg fix for latest version

cawk:
Fixed the code for the latest version released


cawk:
Sorry dont use github pull requests often what are the conflicts?

panthernet:
Means that repo has new commits that are conflicting with your changes. You have to pull these new commits and resolve conflicts locally. Then push your changes again - They will be reflected in this pull automatically.

cawk:
ah i understand as fast as i put my code in there is a new commit which messes with my conflict and as far as i go with these commits the main code is there im sure would only take a minute for someone to fix but i dont have time to keep fixing everytime theres a commit

Issue #793, opened 2016-07-27
[Request] Do not login with default location on the first run


Anyrainel:
I just tried the release v1.0.7. It is nice to have some prompts to input credentials, I thought this program has new feature to configure all details on the first start. However, it directly runs even knowing there was no config.json and auth.json. Could you please not let it run with the default location? It instantly makes me soft-banned.

If the only legit behavior is run once then close, there is no need to input credentials on the first start. You know there are more than account info to configure.

MythHack:
Edit config file directory config.json

MythHack:
"DefaultLatitude": 22.279565983835251,
  "DefaultLongitude": 114.18313980102539,

Longitude and latitude

Anyrainel:
@MythHack Thanks. I know this. I just thought the new release has new features, but it asked for credentials without warning then directly login to the game with the default location.

Anyrainel:
I guess prompt a warning and stop the bot upon creating json files should be reasonable.
Well, nvm if it is hard to change the code. I'll never input credentials in the program again.

smiLLe:
related to #683

Anyrainel:
Okay, People on discord said this 1 sec response time (and so the potential to soft-ban) is how this bot designed.

ShawnHardwick:
I am reopening this issue as it is a common occurrence that the developers need to address.

PR #1073, opened 2016-07-27
Asks if you want to continue on first start

Nordiii:
Just checks if the config exists
yes -> continues normally 
no -> asks if you want to continue with the default config

#793


Diablosito33:
Ahahah! Idk how GitHub works, but this will be added on next update?

PR #1097, opened 2016-07-28
Update for Greek translation.

xakersas:


xakersas:
what should i do?

5andr0:
update your codebase or delete your rep and refork again. then overwrite the current gr.json and please add the latest entries

NecronomiconCoding:
conflicts

PR #1099, opened 2016-07-28
fix renaming pokemons

hakkotsu:
since we are keeping pokemons that are IV >= KeepMinIvPercentage so we should rename them as well


hakkotsu:
integrated in #1115

PR #778, opened 2016-07-27
Bugfix - Incense encounters use pokemon location, not fort location

Ilses:
(Thanks to @Chris#6903 for finding this)

PR #1122, opened 2016-07-28
Refactor CatchPokemonTasks.cs execute function & fix incense.

Ilses:
Migrated from #1026.

Fixes 3 bugs with catching incense Pokemon described in PR #778.

Updated to fix conflicts due to ctx -> session change.

Note that as of present, LureInfo is broken upstream.


yamashi:
Conflicts

PR #1157, opened 2016-07-28
Fix Incense Pokemon Catch Task

emoisback:
Fix error when catch incense pokemon, fortdata is null.


myo:
nice find

csutcliff:
This fix (along with some improvements) has been PR'ed a number of times over the last few days, never merged. I don't think they are interested!

panthernet:
Too much pulls & issue reports for small open-source team to handle effectively. Have to live with it.

NecronomiconCoding:
@csutcliff there you go.

csutcliff:
thanks! :+1:

Ilses:
FML I'd submitted this fix three times. Thanks for helping push this through.

PR #1202, opened 2016-07-28
Optimal default config - Update Settings.cs

tr0n0s:
Optimal config for most people
(I've tested quite bot before suggesting this. and I have discussed with my friends who use. Feel free to add changes and/or discuss it)


tr0n0s:
We aren't programmers, but we will continue working on do better optimal config in teamspeak :D

PR #1249, opened 2016-07-28
FEATURE: randomized delay between each action

Cyberek:
FEATURE: randomized delay between each action

DelayBetweenPlayerActions (Value)
Delay in miliseconds between any actions (such as attempt to catch a
pokemon, evolving a pokemon, farming a pokestop, recycling an itep etc).
This value is randomized between + and - 30% of given value after each
action. If set to 0, default delays are applied (for some of the
actions).


Cyberek:
@NecronomiconCoding could you please take a look?

PR #1266, opened 2016-07-28
Evol filter Option

acongdevil:
Plz add new option to choose which one can evol which one can not.


yamashi:
Please rebase your changes, this contains way too many commits

PR #1272, opened 2016-07-28
[FIX] For Evolution Logic with LuckyEggs

erkutalakus:
LuckyEgg existence should be checked before "UseLuckyEggsMinPokemonAmount" condition to be able to bypass it.

If there is no LuckyEgg in inventory, there is also no need to check "pokemonToEvolve.Count" value. It should just evolve the pokemon.


hadigoh:
What if the users only wants to batch evolve their pokemons when there's a lucky egg (in order to not "lose" double experience)?

PR #1277, opened 2016-07-28
Unbanned while using GPX Pathing, but still requires pokestop.

emoisback:
Unbanned while using GPX Pathing, but still requires pokestop.


yamashi:
Please rebase your commit, this contains too many merge commits

PR #1285, opened 2016-07-28
Unbanned function while using GPX Pathing, but still requires at least 1 pokestop.

emoisback:
Unbanned function while using GPX Pathing, but still requires at least 1 pokestop to start unbanned function.


yamashi:
Please create a function for it.

PR #1289, opened 2016-07-28
[Feature Request]Minimum Pokemon based on candy 

Shibby7986:
Is it possible to add a feature that updates the minimum number of a Pokemon to keep based on the possible evolutions calculated by the amount of candy and the candy required per evolution.

The idea being that when lucky egg is active there is the correct ratio of Pokemon to candy for maximum evolution spam.

this [spreadsheet](https://docs.google.com/spreadsheets/d/1nfMjKK1RXGkqFe8NSZXm5PHDiUVBkcMDl_C-LYb2KNQ/edit) may be helpful to explain further or pull data from.

Logic:
additionalCandy = currentCandy/candyToEvolve 
posibleCandy = currentCandy+additionalCandy
PosibleCandy/candyForEvolve = Pokemon required
minimumPokemon += Pokemon required


MiaoMFoo:
if the bot can add a function : move to a designated place!That will be even more perfect!
That is,if a location has a boss ,type the latitude and longitude spot to move there ,click the boss in to the fight but donnt catch it(throw the ball) right now,
 back to the farm place, and then start capturing, because it will not be banned.

Shibby7986:
Hey Maio, that sounds useful however also best if you create a separate request.

also: checkout the [snipe feature](https://github.com/NecronomiconCoding/NecroBot/pull/1287) being worked on it may do what you are after

MiaoMFoo:
the snipe feature is really cool！~~unfortunately I do not know programming，so ，only i can do is just await and thank for your jobs！

PR #1296, opened 2016-07-28
[Feature Request] Transfer filter by moves (attacks)

nmask:
Hello, 
Id like to suggest to add feature to PokemonsTransferFilter in settings. 

Idea is to filter pokemons by attacks https://pokemon.gameinfo.io/moves, since they are more important than CP or IV. And attacks also are randomly reset after evolving, so its wise to filter by move only end evolved pokemons. 

Thanks.


Misiu:
@nmask Pull request isn't for feature request!

PR #1299, opened 2016-07-28
Fix DelayingUtils properly

csutcliff:


yamashi:
We are removing async and await so this has no reason to be for us

PR #1328, opened 2016-07-29
Snipe Spots and Pokemon + Rename optimized -> Update Settings.cs

tr0n0s:


kaeus:
Why are we adding sniping spots in the default config? If anything, these should be added to WIKI, not in the config.

PoGo.NecroBot.Logic/Settings.cs
```diff
@@ -108,7 +108,7 @@ public class GlobalSettings
         public bool KeepPokemonsThatCanEvolve = false;
         public bool PrioritizeIvOverCp = true;
         public bool RenameAboveIv = true;
-        public string RenameTemplate = "{0}_{1}";
+        public string RenameTemplate = "{1}_{0}";
```
timelf123: "rename optimized"

Longpc:
Please add end of snipping pokemon when we lost all pokeBall, greateBall or superball.
Current BOT still try snipping pokemon when it not have any more ball
@tr0n0s

kaeus:
I'll be adding this as well in a separate PR.

tr0n0s:
you have reason, closed

PR #1334, opened 2016-07-29
Sniping Changes

kaeus:
Disabling Sniping by default. This should NOT be on by default at all. 
Also added config to stop sniping if Pokeballs fall below configured amount (default 15)

### DIFF
--- a/PoGo.NecroBot.Logic/Tasks/SnipePokemonTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/SnipePokemonTask.cs
@@ -277,7 +277,7 @@ namespace PoGo.NecroBot.Logic.Tasks
             ScanResult scanResult;
             try
             {
-                var request = WebRequest.CreateHttp(uri);
+                var request = WebRequest.CreateHttp(uri.Replace(',', '.'));
                 request.Accept = "application/json";
                 request.Method = "GET";
 
--- a/PoGo.NecroBot.Logic/Tasks/SnipePokemonTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/SnipePokemonTask.cs
@@ -11,9 +11,21 @@ using PoGo.NecroBot.Logic.State;
 using PoGo.NecroBot.Logic.Utils;
 using POGOProtos.Enums;
 using POGOProtos.Networking.Responses;
+using System.Net.Sockets;
 
 namespace PoGo.NecroBot.Logic.Tasks
 {
+    public class SniperInfo
+    {
+        public double latitude { get; set; }
+        public double longitude { get; set; }
+        public double iv { get; set; }
+        public DateTime timeStamp { get; set; }
+        public PokemonId id { get; set; }
+        [JsonIgnore]
+        public DateTime timeStampAdded { get; set; } = DateTime.Now;
+    }
+
     public class PokemonLocation
     {
         public long id { get; set; }
@@ -24,6 +36,12 @@ namespace PoGo.NecroBot.Logic.Tasks
         public bool is_alive { get; set; }
         public int pokemonId { get; set; }
 
+        public PokemonLocation(double _latitude, double _longitude)
+        {
+            latitude = _latitude;
+            longitude = _longitude;
+        }
+
         public override int GetHashCode()
         {
             return this.ToString().GetHashCode();
@@ -31,12 +49,26 @@ namespace PoGo.NecroBot.Logic.Tasks
 
         public override string ToString()
         {
-            return uid;
+            return latitude.ToString("0.0000") + ", " + latitude.ToString("0.0000");
         }
 
-        public override bool Equals(object obj)
+        public bool Equals(PokemonLocation obj)
         {
-            return this.ToString().Equals(obj.ToString());
+            return Math.Abs(latitude - obj.latitude) < 0.0001 && Math.Abs(longitude - obj.longitude) < 0.0001;
+        }
+
+        public override bool Equals(Object obj) // contains calls this here
+        {
+            if (obj == null)
+                return false;
+            
+            PokemonLocation p = obj as PokemonLocation;
+            if ((System.Object)p == null) // no cast available
+            {
+                return false;
+            }
+
+            return Math.Abs(latitude - p.latitude) < 0.0001 && Math.Abs(longitude - p.longitude) < 0.0001;
         }
     }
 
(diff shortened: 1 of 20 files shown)


simplyblue333:
Good call, quick way to lose ultraballs. Remember if you're adding sniping pokemon, list is Case-Sensitive.

TheCrimsonDevil:
What's wrong with sniping?

ionred:
@TheCrimsonDevil Drives up potential for bans based on distance traveled, I'm guessing?

kaeus:
Sniping has a purpose, but by default most people want the bot to level. It should be an Opt-In, not Opt-Out.

The other issue is it uses a LOT of pokeballs, so the config helps you from just teleporting around for no reason since you cant catch anything.

Yodude12:
great how to disable and how to get as many pokeballs be4 catching ?

TheCrimsonDevil:
@ionred My understanding is that sniping avoids the softban by teleporting there for a moment, starting the encounter, then teleporting back, which doesn't get you softbanned since the system just assumes it's a one-off glitch, THEN once you're back in your original spot, it throws the pokeball.

TheCrimsonDevil:
For the record, I do agree it should be opt-out as opposed to opt-in.

ionred:
I guess I didn't understand that piece. Good info to have. Thanks!

LordAlzion:
and maybe add more pokemon option to the sniping list

kaeus:
What do you mean options? I'd prefer to let people configure their lists individually rather than us all doing the exact same sniping.

LordAlzion:
thats basicaly what i mean, but the pokemon location near the user is not always the best for sniping so perhaps adding more location around the world then allowing the user to disable it in settings is better? that way ppl can have the best sniping location for any pokemon they want. or maybe we can just make a list on best sniping location then ppl can add manualy hmm...

kaeus:
Again, as the same with location/GPX we should leave this up to the user. I'm perfectly OK with adding this to the Wiki, but adding it into the bot by default seems like bad practice.

timelf123:
Agreed, considering the majority of people downloading the precompiled release are going to leave the sniping settings as is

PoGo.NecroBot.Logic/Tasks/SnipePokemonTask.cs
```diff
@@ -79,6 +79,14 @@ public static class SnipePokemonTask
...
                         foreach (var pokemonLocation in locationsToSnipe)
                         {
+                            var pokeBallsCount = await session.Inventory.GetItemAmountByType(POGOProtos.Inventory.Item.ItemId.ItemPokeBall);
+                            var greatBallsCount = await session.Inventory.GetItemAmountByType(POGOProtos.Inventory.Item.ItemId.ItemGreatBall);
+                            var ultraBallsCount = await session.Inventory.GetItemAmountByType(POGOProtos.Inventory.Item.ItemId.ItemUltraBall);
+                            var masterBallsCount = await session.Inventory.GetItemAmountByType(POGOProtos.Inventory.Item.ItemId.ItemMasterBall);
+
+                            if (pokeBallsCount + greatBallsCount + ultraBallsCount + masterBallsCount < session.LogicSettings.MinPokeballsToSnipe)
+                                return;
+
```
sliverfishes: Shouldn't this be before the foreach loop so that we can do all the locations and not just 1-2? Or maybe random which location to go to each time?

kaeus: I mean, the point of this is to avoid checking a location if you run out of balls. Worst case, you fail to be able to catch a few at one location and skip the rest until you hit more pokestops.

sliverfishes: That's the problem though, you don't check other locations beyond the first 1-2 with this. It farms balls, gets just a few more than the limit to snipe, starts sniping at the first location, falls below the limit, stops sniping, repeat. It always starts sniping at the same location.

kaeus: The only way to solve that is to randomize which locations to do or to make sure your pokeball count is high enough to make it through all. TBH after a few iterations the first stops will clear out somewhat and you'll make it to the next.

sliverfishes: That's exactly what I said originally...

The dratini spot never clears out for me though, it gets about 10 dratinis every visit.

esme518:
the bot will also farm Pokemons without Pokeball.
so can u also added config to stop farm Pokemons if Pokeballs is zero, 
and change the bot into something like "Farm pokestops" mode until Pokeballs reach configured amount.

rastinder:
can we do smthing "https://twitter.com/RAREpokespawn" this bot always tells the locations of rare pokemons so is it possible to snipe these pokemons automatically ???

KidLoad93:
this should have a config that stop farm pokemon when poke ball is <5 ball and just go like normal.thanks for fix it soon.

tangdami:
Getting rare pokemons from some alert webpage and then sniping them would be a great idea!

Here are three facebook real-time pages you can have a look:
**Snorlax**: https://www.facebook.com/snorlaxspawns
**Lapras**: https://www.facebook.com/LaprasSpawns/
**Dragonite**: https://www.facebook.com/DragoniteSpawns/

me2rabbit:
bug report, when set snipping true and there are great amounts of pokemon ( i.e. Dratini, Eevee) to catch, Transferring, or possibly evolving is not working which results in PokemonIventory full.

[00:29:41] (CatchSuccess) | (Normal) Charmander Lvl: 14 CP: (326/382) IV: 38.00% | Chance: 16% | 8620293.73m dist | with a GreatBall (9 left). | Candies: 64
[00:29:51] PokemonInventory is Full.Please Transfer pokemon manually or set TransferDuplicatePokemon to true in settings...
[00:29:54] [Sniper] Scanning for Snipeable Pokemon at 38.5568074864611,-121.238379478455...
......
[00:35:14] PokemonInventory is Full.Please Transfer pokemon manually or set TransferDuplicatePokemon to true in settings...
[00:35:16] [Sniper] Scanning for Snipeable Pokemon at 37.825329,-122.464897...
[00:35:16] [Sniper] No Pokemon found to snipe!
[00:35:17] Eevee	- CP: 339  IV: 62.92%   [Best CP: 458  IV: 96.52%] (Candies: 16)

As seen above, snipping force to loop every location even when inventory full and unable to perform transferring

Brumotti:
the idea is good sniper.
Add to it a function, for example to make an interactive configuration sniper.json (to, without switching off the bot, he would pick up a real-time value of a config) and enter the coordinates into it in such a way, only a single line:
30.450943316119, -84.273267563384
with a rare Pokémon obtained with pokevision, in real time.

ie found Pokemon introduced coordinates and saved, the bot teleport and caught him. You can add a few attempts, and then config cleaning bot to not constantly jumping on those coordinates.

warkantos:
I´m using this site 
http://pkmngotrading.com/wiki/Pokemon

It has the coordinates of several nests for use with snipping.

jaskovic:
guys why im getting this ??
[14:53:56] (INFO) [Sniper] Scanning for Snipeable Pokemon at 38,5568074864611,-121,238379478455...
[14:53:58] (INFO) [Sniper] No Pokemon found to snipe!
[14:53:58] (INFO) [Sniper] Scanning for Snipeable Pokemon at -33,859019,151,213098...
[14:53:58] (INFO) [Sniper] No Pokemon found to snipe!
[14:53:58] (INFO) [Sniper] Scanning for Snipeable Pokemon at 47,5014969,-122,0959568...
[14:53:58] (INFO) [Sniper] No Pokemon found to snipe!
[14:53:58] (INFO) [Sniper] Scanning for Snipeable Pokemon at 40,785091,-73,968285...
[14:53:58] (INFO) [Sniper] No Pokemon found to snipe!
[14:53:58] (INFO) [Sniper] Scanning for Snipeable Pokemon at 51,5025343,-0,2055027...
[14:53:59] (INFO) [Sniper] No Pokemon found to snipe!

yamashi:
Please get rid of the merge commits, use pull rebase

PR #1338, opened 2016-07-29
auto start and continue without pressing any keys

fee622:
Always getting Error at night and stopped, can it change  to auto restart without pressing any keys or asking?


reflexx666:
"StartupWelcomeDelay": false, and it will

fee622:
thanks

PR #1339, opened 2016-07-29
Added check for CurrencyDecimalSeparator

Brndstrpvj:
Check CurrencyDecimalSeparator to make sure "." is used as seperator in lat and long when calling pokevision.com.


lyrex:
Oh well, I haven't seen your commit. Now I created a similiar solution, but probably a bit more on the clean side.
Check out #1340 which correctly changes locale settings globally instead of just replacing commas with dots.

Brndstrpvj:
This is my first experience with non-script languages so I actually appreciate seeing a proper way. I did wonder whether changing the locale settings globally would be overkill since it only makes a difference when using pokevison

lyrex:
Depends on how you see it. I think it's not that intrusive that it's an overkill.
Especially if you think about the future. The same problem can occur again if it's used again in an other part in the future which would not happen if the locale is set correctly.

PR #1344, opened 2016-07-29
Fix translation file for French and Norwegian

lyrex:


Colwyn29:
There is a mistake on the code for fr. Check at the end and you will see something like this : 
},
,
{

Delete the ,

PR #1346, opened 2016-07-29
Functionality to transfer amost all (or all) pokemons at fully evolved state

domints:
As for now, this bot might leave you with, for example, 30 of Hypnos, because they have high CP and cannot be evolved anymore. With this change, you would be able to leave only few of them and transfer the rest.
Settings for this allow you to turn it off and on, and set how many pokemons should be left.
Of course you can set how much pokemons should be left, individually for each of them, too.

### DIFF
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -47,16 +47,18 @@ namespace PoGo.NecroBot.Logic
         {
         }
 
-        public TransferFilter(int keepMinCp, float keepMinIvPercentage, int keepMinDuplicatePokemon)
+        public TransferFilter(int keepMinCp, float keepMinIvPercentage, int keepMinDuplicatePokemon, int keepEvolvedDuplicates)
         {
             KeepMinCp = keepMinCp;
             KeepMinIvPercentage = keepMinIvPercentage;
             KeepMinDuplicatePokemon = keepMinDuplicatePokemon;
+            KeepEvolvedDuplicates = keepEvolvedDuplicates;
         }
 
         public int KeepMinCp { get; set; }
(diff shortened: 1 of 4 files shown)


Plawed:
It also seems the bot does not at all transfer duplicates of: Weezing, Parasect or Arcanine... possibility of others. Maybe the PokemonID enum IDs arent there? Wish that was back in the code directly... I think those IDs are all by included DLL now.

domints:
Maybe because they match the non-transfer pattern. I haven't seen it transfer even hypno. And this PR allows you to set this bot to fix it. My change ignores other transfer-specific changes, if it matches.

kingchenc:
Thx, this fix work for me for evolved pokes, but can we have this fix too for non'evolved pokes? :)

domints:
You have lots of unevolved pokemons left? Weird. I could think of that, when I'm home :) So this should override other transferring settings, leaving you with desired amount of any pokemon? I should then check the order of tasks :)

domints:
I have question for you, because your config was probably pure. What values did "KeepEvolvedDuplicates" get in "PokemonsTransferFilter" config session? Is it 0 or 2?

kingchenc:
Hello, @domints Yes, 90% of my Pokes evolve and the rest not.. i dont know whats the problem, im not the pro here, in the log i see no error or so... 
and the value that i use is 2

ShawnHardwick:
Closing due to merging conflicts. We are preparing for a new release so please resubmit if the implementation is still relevant for the new release and also when the conflicts have been handled 

Secondly, this can be solved with the PokemonTransferFilter as well.

domints:
Merging conflict came just not long ago :) I will solve it, if it will be needed.
And, about secondly, PokemonTransferFilter works for single pokemon type, while my mod for all genres.

PR #1348, opened 2016-07-29
Update translation.nl.json

WinSCaP:
Small updates to make the Dutch translation more correct.


deknarf:
I don't think the names of PokeBalls change, do they?

WinSCaP:
@deknarf TV series says so :) (My nephew is a real mudkip fan)

PR #1362, opened 2016-07-29
Throw away low level pokeballs, potions, revives and keep the better ones.

th3ragex:
Hi,
i changed the recylce behaviour to throw away low level pokeballs and potions and keep the better ones. This leads to a far more frequent usage of the better pokeballs instead of throwing them away. The new setting entires configure the total amout of pokebals/potions over all types that shall be kept.

### DIFF
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -90,6 +90,8 @@ namespace PoGo.NecroBot.Logic
         string ProfileConfigPath { get; }
         string GeneralConfigPath { get; }
         bool SnipeAtPokestops { get; }
+        int TotalAmountOfPokebalsToKeep { get; }
+        int TotalAmountOfPotionsToKeep { get; }
 
         ICollection<KeyValuePair<ItemId, int>> ItemRecycleFilter { get; }
 
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -97,6 +97,7 @@ namespace PoGo.NecroBot.Logic
         int MinDelayBetweenSnipes { get; }
         int TotalAmountOfPokebalsToKeep { get; }
         int TotalAmountOfPotionsToKeep { get; }
+        int TotalAmountOfRevivesToKeep { get; }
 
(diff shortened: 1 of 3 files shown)


nfettinger:
This is a duplicate.

KicksDNA:
Is there any way to keep old auth and config files when the bot auto updates??

th3ragex:
So this got already fixed by someone else?

Shibby7986:
In Config you can control the levels.
Change "itemPokeBall"'s value to 0 and it will transfer them all.

`"ItemRecycleFilter": [
    {
      "Key": "itemUnknown",
      "Value": 0
    },
    {
      "Key": "itemPokeBall",
      "Value": 25
    },
    {
      "Key": "itemGreatBall",
      "Value": 50
    },
    {
      "Key": "itemUltraBall",
      "Value": 100
    },
    {
      "Key": "itemMasterBall",
      "Value": 100
    },
    {
      "Key": "itemPotion",
      "Value": 0`

th3ragex:
@Shibby7986
I know, but that is not what i wand and my codes does. 

Let me try to explain it. Lets say i configure the bot with the settings you posted. The bot will always throw away greatball until 50 remain, and ultraballs until 100 remain. It will additionally prefer using all normal pokeballs until they are empty. (except a high CP or hard to catch pokemon appreads.)

I changed it to keep, lets say, 100 pokeballs. If i am starving on greatballs the normal pokeballs will be kept, if i get a lot of better pokeballs the bot will keep them and throw the lower pokeballs away. This also reduces the amount of different items which the bot has to throw away => faster.

Changing the config to throw away all lower pokeballs may cause problems if you don't get them from the pokestops.

Ryuichiro:
Looks great, is there a reason you didn't use the same logic for Revive and MaxRevive?

th3ragex:
@Ryuichiro good point, never saw a MaxRevive yet. I will add it right now.

Suslikuraip:
Also recycled counter shows how many times recycling was made not actual amount of stuff thrown away.

SindreKjetsa:
I was gonne edit the settings to something similar, very good commit :) hope it's commited soon

Ryuichiro:
@th3ragex I have some questions about Git Pullrequests, could you write me in Slack. Same name: Ryuichiro

Dano7:
Cool change!

Ryuichiro:
@th3ragex Do you even use Discord?

PR #1367, opened 2016-07-29
Sniping Updates

kaeus:
Disabled Sniping by default.
Setting to stop sniping if under min pokeball amount (20 default)


solenrus:
can you add an extra check to initialize the transfer when the pokemonInventory gets full?
cause I get this all the time https://gyazo.com/ebaaafb947167a70378db62492d666ce
cheers
@kaeus

PR #1371, opened 2016-07-29
use ultraball after pokeball over

roi89:
the bot start using ultraball after my normal pokeball has over and the bot take all ultraball and still trying to catch pokemon and the bot say that isnt enough pokeballs to catch the pokemon and the bot isnt trying to go to other pokestop to get pokeball and still trying to catch pokemons,


roi89:
can you explain me what to do?

PR #1373, opened 2016-07-29
update google auth

mobab:
There is a problem with the new authentication way, when it was using only the token it was working,
Now using the username and the password, it is saying:

> As you have Google Two Factor Auth enabled, you will need to insert an App Specific Password into the auth.json

And in fact i dont have such setting, and the webpage it is opening reports

> The setting you are looking for is not available for your account.

My google account is managed by my university account. what is the solution for this, gmail that belongs to loughborough university.
error1
error2


kea0811:
Is this fix yet? Can we do it w/o app password?

PR #1388, opened 2016-07-29
Added output alerting user they have no pokeballs

bornsupercharged:
Previously the code would exit without notifying the user that they are not catching any pokemon because they have no pokeballs.


NecronomiconCoding:

PR #1389, opened 2016-07-29
(INFO) Program will continue after the key press...

PokemonRS:
problem solved


lotusJDM:
Having the same issue 
image

PR #1390, opened 2016-07-29
Fixed MinPokeballsToSnipe

Ryuichiro:
Now the Bot will snipe until he's out of pokeballs and tell you why he
stopped or didn't start.

I never worked with CancellationToken so please double check the code.

PR #1394, opened 2016-07-29
Pokeballs check for Feeder & IV Logic fix

5andr0:


Ryuichiro:
I'm irritated, how can you add a logic now, that I fixed an hour ago? oO
https://github.com/NecronomiconCoding/NecroBot/pull/1390/files

Also I have no idea how the IV logic works, why check ">0" if you check ">KeepMinIvPercentage" as well?

Dano7:
Nice.  I implemented both the snipe filtering and the pokeballs (feeder) check locally today too.  But your snipe filter is a lot cleaner by just reusing the transfer filter.

It might be nice to add a setting to decide whether to use ultraballs and masterballs for sniping, as sniping will easily use up all your balls capturing low IV Eevees.  That way you don't have to use a strict filter, you can use up all your pokeballs and greatballs sniping whatever is new in the area, but stop sniping when you are out of pokeballs or greatballs.   

I look at sniping as bonus way to use up my pokeballs before/rather than recycling them.   It can be really hard to get a filter that would balance the benefit of sniping with wasting ultraballs low level Eevees, unless you also can balance it against available ammo.   

It would be a shame to recycle balls because your filter was a little too tough at the moment, but you also don't want to waste ultraballs..??   Just an idea.    Maybe the setting could be Sniping ammo: and specify be any of the Ball types, and sniping can keep going until all of the specified ball types are gone, or no more pokemons match the filter.

5andr0:
Sry @Ryuichiro didn't see. About the IV: forgot to remove IV > 0 logic should've been IV > 0 || IV > monster_iv to catch unknown iv mobs

PR #1402, opened 2016-07-29
Important Snipes

mich3lam:


ifbatmanwasreal:
Does this work? Have you tested? Because when I add too many to the snipe list it doesn't work correctly.

mich3lam:
@ifbatmanwasreal lol yes it works

ShallowHacker:
How do I input? It looks different lol

ifbatmanwasreal:
He changed the way they are identified. Just follow copy his and replace it with your old. And then follow his format if you add Pokemon to the list.

That's the quickest way.

Yodude12:
he meant which file should he put these new pokemon in since we were only using the notepad to edit ... right ??

ShallowHacker:
Nah, I know.

Right into the config with Notepad++

PR #1404, opened 2016-07-29
Improved Sniper IV logic

5andr0:


mich3lam:
So this doesnt snipe if the IV isnt shown on the channel it grabs the info from ?

PR #1412, opened 2016-07-30
farm pokemon

lnkakalink:
make points selection function for farm Pokemon
. Statistics also do in places where there is information about how the site around the points and so on.
just lay out their coordinates to farm here

### DIFF
--- a/PoGo.NecroBot.Logic/PoGo.NecroBot.Logic.csproj
+++ b/PoGo.NecroBot.Logic/PoGo.NecroBot.Logic.csproj
@@ -42,6 +42,7 @@
     <Reference Include="System.Configuration" />
     <Reference Include="System.Core" />
     <Reference Include="System.Device" />
+    <Reference Include="System.IO.Compression.FileSystem" />
     <Reference Include="System.Xml.Linq" />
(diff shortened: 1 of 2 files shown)


lnkakalink:
"DefaultLatitude": -33.8731667,
  "DefaultLongitude": 151.2082071,

Fenixxia:
Dear developers. You have done a lot. But still not entered one of the main filters to pokemon! Check for a match "Type of pokemon" = "attack type pokemon". This correspondence is called STAB. Attack pokemon gets a bonus of 1.25. Pokemon without STAB loses to pokemon with STAB at the same LV, CP, and IV of about 25-30%. 
Now collect all the pokemon based on the priority IV, but why is it needed without STAB?
Because the type of attack the pokemon can change during the evolution of such compliance testing need to take place for 2 and 3 levels of evolution. 
With great hope for the early entry of such functions.

Ilses:
@Fenixxia This is a pull request, not a feature request. Save your entry for the return of the issue tracker, please.

Fenixxia:
I have no way to make a request to add new features through Github

Ilses:
You're hijacking a PR. PM a dev, or hold off on the request for now.

PR #1415, opened 2016-07-30
Add settings for throw accuracy and spin to humanize pokemon catching

scristall:
After some testing, I found that the CatchPokemon API's final 3 parameters seem to do the following:

normalizedRecticleSize: value from ~0.1 to ~1.95 which determines the quality of the throw.  Low values will have high escape rates, middle values will trigger a "Nice" catch, and high values a "Great" or "Excellent" one.

spinModifier:  Although it's a double, it appears to be a boolean.  If this value is 0, the ball will not have a spin bonus, otherwise it will.

normalizedHitPos:  Can't seem to figure out what this one does -- any value other than 1 triggers CatchError as far as I can tell, but needs further testing.

This pull request uses this information to allow for setting these values, and defaults the accuracy of the throw to 0.5 to 1.95 with a uniform distribution.  It also sets a 50% rate for the Spin Throw.

This pull will (by default) drastically reduce Exp gain from Pokemon catching, but will cause the bot to look less like a bot.

### DIFF
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -101,6 +101,9 @@ namespace PoGo.NecroBot.Logic
         int TotalAmountOfPokebalsToKeep { get; }
         int TotalAmountOfPotionsToKeep { get; }
         int TotalAmountOfRevivesToKeep { get; }
+        double ThrowMaxAccuracy { get; }
+        double ThrowMinAccuracy { get; }
+        double ThrowSpinRatio { get; }
 
         ICollection<KeyValuePair<ItemId, int>> ItemRecycleFilter { get; }
 
--- a/PoGo.NecroBot.Logic/Settings.cs
+++ b/PoGo.NecroBot.Logic/Settings.cs
@@ -142,6 +142,9 @@ namespace PoGo.NecroBot.CLI
         public int TotalAmountOfPokebalsToKeep = 150;
         public int TotalAmountOfPotionsToKeep = 100;
         public int TotalAmountOfRevivesToKeep = 50;
+        public double ThrowMaxAccuracy = 1.95;
+        public double ThrowMinAccuracy = 0.1;
+        public double ThrowSpinRatio = 0.3;
 
 
         public List<KeyValuePair<ItemId, int>> ItemRecycleFilter = new List<KeyValuePair<ItemId, int>>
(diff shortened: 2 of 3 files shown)


septemus:
I don't feel like this is needed (at least, yet). We're all as good as banned so I don't feel like humanizing features are really necessary at this point.

norecha:
If you don't feel like you need it, you won't use it. I'm sure many users, including me, would like to have this feature.

hilycker:
@septemus You're right, we're all as good as banned, but sooner or later the aim will be to humanize the bot as much as possible, otherwise why use it? And it's also an improvement, so thumbs up

boullose:
What makes your sure everyone who uses the bot will get banned?

DrWhatNoName:
@boullose ever played Niantics other game? Ingress? that game was botted just the same at pokemon go is now. Everyone who used a bot/GPS spoofer got banned.

jobber2063:
Niantic decided it doesn't like my wifi connection. Can't login with any account (bot used or not) and necrobot isn't connecting either. Phone works fine on 4g... No idea how to explain that.

Ilses:
This represents a fairly drastic change to the current behavior, I would recommend adding a toggle for it. Probably a config option "Humanize" : true/false.

scristall:
@Ilses Maybe setting the default behavior to be perfect throws like it is now would be better?  A blanket Humanize setting is somewhat heavy handed, since more humanizing algorithms may be written down the line.

GittyHubb:
how do you make it 100% excellent and spin all the time?

scristall:
@GittyHubb You would set ThrowMinAccuracy to 1.95 and ThrowSpinRatio to 1.0.  I suppose we could map these to more understandable values as well.

Cyberek:
Sounds fine. maybe this "humanize" way could cover more aspects (randomized delays between actions which is already implemented in some parts, throw accuracy etc)

ModernTimes:
+1 for more human behavior!

- Default: perfect throws
- Option to toggle humanized throws
- with additional parameters to specify the bell curve for accuracy
- Default options for these additional parameters should lead to a small chance to miss the pokemon altogether

NecronomiconCoding:
I dont think this is needed right now. Once we see some changes in the way Niantic Detects things we can merge this.

Emuuung:
This was a nice way to double up EXP by catching pokemon every once in a while, doesn't need to be "always perfect throw".

PR #1325, opened 2016-07-29
Rename all pokemon with IV that aren't already nicknamed.

Sekioh:
Adds another configuration option that renames anything that's not custom named, to be another choice over the renaming only above capture IV minimums.

PR #1416, opened 2016-07-30
Rename all pokemon with IV that aren't already nicknamed.

Sekioh:
Resolved conflicts for #1325.

### DIFF
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -84,6 +84,7 @@ namespace PoGo.NecroBot.Logic
         float EvolveAboveIvValue { get; }
         bool DumpPokemonStats { get; }
         bool RenameAboveIv { get; }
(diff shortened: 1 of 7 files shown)


solenrus:
great stuff!

DeadElle:
approved +1

sub7even:
May you can add an option to rename all Pokemon (and not only the ones that aren't already nicknamed)?

I'd  like to change the Naming format and in this case I need to update all Pokemon (including the one who already have a nickname).

Sekioh:
Naming format is already supported by the newer revisions under the translation files. I'll consider changing the flag from toggle to a string like "All", "Original", "MinimumIV". Then people can pick from options.

Ryuichiro:
I created a very similar solution, but I think it should be easier to extend and I added a skip for favorites.
[#1431](https://github.com/NecronomiconCoding/NecroBot/pull/1431)

NecronomiconCoding:

PR #1417, opened 2016-07-30
FarmPokestopsGPXTask completely broken, iterating only through the fi…

JaroslavKaas:
…rst segment of the first track over and over again


giesti:
Correct, this loops infinite. Also should have an option to run the track once then stop.

PR #1422, opened 2016-07-30
Updated German trans to show Ä,Ö,Ü as AE,OE and UE

schadis:
Removes the bug from showing wrong characters in cmd


Epinikion:
Theres no need for that i guess, u can change the used font in cmd to e.g. "Consolas" which will show the right Umlaut even in Pokestop names.

right click cmd title -> settings ->  font -> choose "Consolas" (I use Consolas at 14pt)

RoseFlunder:
Yes @Epinikion is right. I also you "Consolas" as font and all characters are displayed correctly.

schadis:
k ty

PR #1424, opened 2016-07-30
Min delay between Item Recycle & Min KM to walk for limited incubator

ldevil-ch:
I kept getting recycles after almost every pokestop so I added a minimum delay to avoid waste of time. Recycling bigger stacks saves time.

I also was seeing 2km eggs being put into my limited use incubators so I put an according line to prevent the bot from using small eggs. Can be set in settings (2 = all, 5 = skip 2km, 10= skip 2 & 5 km and 11+ = don't use limited use incubators)

This is my first code change so please check and constructive feedback is always welcome.


NecronomiconCoding:

PR #1429, opened 2016-07-30
Added Danish translation file [Fixed with .csproj]

FL4ZH:
Translated the German translation file into a Danish translation file,
corrected the directory and added it to .csproj


amivit:
Sådan, godt lavet :)

PR #1430, opened 2016-07-30
[add] Never spawn at the same position. Niantic might detect that!

Floppixx:
Dont know if they do it. But better safe than sorry!

### DIFF
--- a/PoGo.NecroBot.Logic/Settings.cs
+++ b/PoGo.NecroBot.Logic/Settings.cs
@@ -458,11 +458,14 @@ namespace PoGo.NecroBot.CLI
             }
         }
 
+        // Never spawn at the same position.
+        private readonly Random _rand = new Random();
+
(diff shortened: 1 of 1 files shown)


Jackman1000:
Yeah, suddenly, im not spawning in the same location as the one in my .ini

stainx:
Would it be possible to cache the bots last known position for the next time it loads? With an manual override option for new locations.

mich3lam:
this commit ruins the bot, it will always start somewhere and then it';ll need to walk back..

Floppixx:
Oh yes. Increse the offset.

Zuxser:
Nice idea, bu on equator 0,2 deg in each direction might be up to 30 km from origin place (max on equator).
untitled

Floppixx:
Yes. 0,2 is to much. Never got that on my tests. Should be max 0,05

PR #1431, opened 2016-07-30
Enhanced Renaming

Ryuichiro:
1. The bot won't rename favorites.
2. Instead of "RenameAboveIv" you now have
"RenamePokemon" which renames all pokemon not transfered and
"RenameOnlyAboveIv" which skips pokemon with less than
KeepMinIvPercentage IV.

With both true you have the old behavior.
"RenameOnlyAboveIv" alone will do nothing.

I think it's easier to maintain and extend than [#1416](https://github.com/NecronomiconCoding/NecroBot/pull/1416)


solenrus:
perfect, thanks!

PR #1437, opened 2016-07-30
Event: send current position on start

nicoschmitt:
so we can show the map before all the cleaning in inventory that takes some time


Ryuichiro:
Maybe comment the call, so we will remember why we call it? ^^

PR #1448, opened 2016-07-30
Added pokeball output for counts and an alert if they have none

bornsupercharged:
Added pokeball output for counts and an alert if they have none while trying to catch a pokemon.


NecronomiconCoding:
translations missing

bornsupercharged:
Is this how I reopen?

PR #1450, opened 2016-07-30
Added pokeball output for counts and an alert if they have none

BornSupercharged:
Has translations now; added pokeball output for counts and an alert if they have none while trying to catch a pokemon.


bornsupercharged:
This should fix the old PR

bornsupercharged:
All translations in place now

ramsedsay:
Right on! Now is this from your own difference.?

PR #1456, opened 2016-07-30
Offset parameter

Ryuichiro:
Fixed version for randomized spawn location.
[#1430](https://github.com/NecronomiconCoding/NecroBot/pull/1430)


Ryuichiro:
Now based on position and in meters.

Zuxser:
Did you tested it?
_settings.MaxSpawnLocationOffset is type of int? So you dividing int with an much more larger int - does it cause fraction loss and 0 as a result? 
return _settings.DefaultLatitude + _rand.NextDouble() * (_settings.MaxSpawnLocationOffset / 111111);
return some value + (random value * 0) -> return some value

Ryuichiro:
@Zuxser Not sure, it worked so I didn't take a closer look. But I will make a fixed Pull request to be sure.

Ryuichiro:
[#1461](https://github.com/NecronomiconCoding/NecroBot/pull/1461)

PR #1457, opened 2016-07-30
not needed anymore

Happymeal1887:


Oolaruu:
So we can delete those strings?
"openingGoogleDevicePage"
"couldntCopyToClipboard"
"couldntCopyToClipboard2"

Any other that are not used anymore?

Happymeal1887:
i thing thats all for now

PR #1461, opened 2016-07-30
Offset parameter explicitcast

Ryuichiro:
Explicit double cast, not sure if nessecary?


Zuxser:
Why just not making _settings.MaxSpawnLocationOffset double type?

PR #1462, opened 2016-07-30
TotalAmountOfPokebalsToKeep lowered to 100

DeadElle:
• if account has stacked incenses/lucky eggs/lures since level1 would exceed 350 items limit and trigger the full inventory continuosly


NecronomiconCoding:

PR #1470, opened 2016-07-30
LogInventory realtime json dump feature

reuscam:
Toggling this allows the bot to dump the entire raw inventory to disk, for consumption by various third party tools.  One use would be an inventory view.

Dump path: /temp/{playername}.inventory.json

New Settings:
LogInventory (bool) - default false
LogInventoryDelaySeconds (int) - default 30


nicoschmitt:
Why do you need to dump inventory to file every x seconds?
That's what gui are for

reuscam:
The point is to decouple the UI from the engine.  This will easily allow third parties to provide inventory tools, without modifying the necro solution.

nicoschmitt:
That is what websocket is there for. Parsing log file should not be the way.
But that's my opinion, so the PR may get accepted anyway :)

reuscam:
Reasonable point.  Maybe I can get this PR in, then refactor to websocket in a second PR.  Would really like to get this one to stick so I can keep my inventory tool working, and stop manually merging settings.cs.

PR #1474, opened 2016-07-30
Upgrade fix

l3uddz:
Wasnt upgrading when absolute value was specified, e.g. wanting only
100% IV uprades.


poplove:
I am a Chinese, I admire you very much, so I use the translation software to communicate with you.
Why a new version every time you start to press any key to continue, it will not automatically reconnect midway dropped. If you sleep at night, the script will stop there. Hope to solve this problem

PR #1475, opened 2016-07-30
Update translation.ru_RU.json

bizideas:
In Russian language absent word "Итемов", there is more appropriate in this context, the word "Предметы"


Fenixxia:
In this case we used the gaming lexicon. Second, such writing saves space, allowing to better perceive the information.

DurtyFree:
See @Fenixxia comment.

PR #1477, opened 2016-07-30
Added config option: "KeepEvolvablesOnlyIfOnEvolutionList"

GammaPaladin:
Added config option: "KeepEvolvablesOnlyIfOnEvolutionList"
Defaults to false.
If "KeepPokemonsThatCanEvolve" is true, and
"KeepEvolvablesOnlyIfOnEvolutionList" is also true, the bot will NOT
keep pokemon it has enough candy to evolve, unless they are on the
"PokemonsToEvolve" list.


rherartt:
this will be very useful!

NecronomiconCoding:
fix conflicts :smile:

PR #1481, opened 2016-07-30
Upgrade all pokemon above IV % OR CP and use PokemonToLevelUp list

tehaon:
The previous code was all sorts of broken for and even when I did finally get it to work it only upgraded like 1 pokemon once... so I looped it and made it get all pokemon above IV instead of just from your top list

### DIFF
--- a/PoGo.NecroBot.Logic/Tasks/FarmPokestopsTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/FarmPokestopsTask.cs
@@ -48,6 +48,7 @@ namespace PoGo.NecroBot.Logic.Tasks
 
             var pokestopList = await GetPokeStops(session);
             var stopsHit = 0;
+            var displayStatsHit = 0;
             var eggWalker = new EggWalker(1000, session);
 
             if (pokestopList.Count <= 0)
@@ -158,6 +159,9 @@ namespace PoGo.NecroBot.Logic.Tasks
                 if (++stopsHit%5 == 0) //TODO: OR item/pokemon bag is full
                 {
                     stopsHit = 0;
+
+                    await DownloadProfile(session);
+
                     if (fortSearch.ItemsAwarded.Count > 0)
                     {
                         await session.Inventory.RefreshCachedInventory();
@@ -179,6 +183,11 @@ namespace PoGo.NecroBot.Logic.Tasks
                     if (session.LogicSettings.RenamePokemon)
                     {
                         await RenamePokemonTask.Execute(session, cancellationToken);
+                    }
+                    if (++displayStatsHit >= 4)
+                    {
+                        await DisplayPokemonStatsTask.Execute(session);
+                        displayStatsHit = 0;
                     }
                 }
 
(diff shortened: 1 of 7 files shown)


solenrus:
make the same for CP please, thank you.

skupfer:
While you are on it, gladly I looked over the current PRs,

Could you add a PokemonToLevelUP list?

```
        public List<PokemonId> PokemonsToLevelUP = new List<PokemonId>
        {
            PokemonId.Dragonite,
            PokemonId.Gyarados
        };
```

So only Pokemons on this list get upgraded?

Or maybe we need another config value which lets users decide whether they want to upgrade ALL pokemons fitting the values or only those on the list fitting the values...

If I PR now, it will have conflicts and I know how lucky I am, as soon as I resolve them, Necro is away and I get new ones :D

solenrus:
We just need something that would lvl up by IV or CP all the way to max lvl, not just once.
A list would be nice too.

DilatoryCheetah:
You can actually give a meaningful error message based on the possible response result:

                ErrorUpgradeNotAvailable = 4,
                ErrorPokemonIsDeployed = 5

Those should be if the pokemon is at max level for your player level and if the pokemon is in a gym, respectively.

Come to think of it, you should also exclude any pokemon that are in gyms from the list of pokemon to upgrade

tehaon:
Is it sad that I like all of these ideas you guys suggested but I don't know how to add more changes to this request? lol I can't even figure out how to access my patch-1 branch :(

tehaon:
figured it out. working on implementing requests.

tehaon:
All requests implemented

cawk:
Wow brilliant additions to the upgrade feature great to see it improved rather than just left as it was when i made it

tehaon:
We appreciate your work to add the initial feature.. I wish there was a good way to keep it from spamming the error messages at every stop though when its out of candies or is max level.

bmdr38:
@tehaon feel free to grab from my pr as well and merge into this for all the additional checks (stardust/candy/level reqs, and everything related to those tasks)
Let me know if thats the case, cause then ill cancel my pr and then everyone gets it in one go instead of 2

tehaon:
I added in your code bnmdr38 with a few changes. Look over it when you get the chance.

PoGo.NecroBot.Logic/Tasks/FarmPokestopsTask.cs
```diff
@@ -208,5 +217,14 @@ public static class FarmPokestopsTask
 
             return pokeStops.ToList();
         }
+
+        private static async Task DownloadProfile(ISession session)
```
bmdr38: Personally, I think this should have the comment on why it was copied from LoginState and made static, since otherwise it purely looks like it is duplicated (aside from the fact that it is static here and not in LoginState)

bmdr38:
Looks good to me, I only had one minor nitpick.  Thanks for taking this up

tehaon:
what the hell happened to my request... now it says I have 236 changes? lol

bmdr38:
Looks like history might have gotten screwed when I closed my pr?  That doesnt exactly make sense to me, but its the only thing I can think of.  Maybe a rebase will resolve?

NecronomiconCoding:
fix conflicts :smile:

PR #1483, opened 2016-07-30
Add choice of how many times to upgrade

cawk:
Can choose how many times they want to loop through the pokemon upgrades
if they hit a resource error or max pokemon level error it breaks loop to stop wasting time


cawk:
Also Fixes an exception bug with the max level

var1on:
Perhaps you should put the rnd generation out of the for loop or remove the breaks. This way the bot will lvl up different pokemon but will stop the loop if one is max level

ShawnHardwick:
May be a duplicate of #1481

PR #1487, opened 2016-07-30
Bot was catching all pokemons above 300 with greatball

Revell1on:
Removed a code which makes bot to use greatball on ANY pokemon above 300 CP


80rrego:
Same issue the account got 5 pokeballs and is using greatballs for any pokemon, the list to not transfeer is not working for my Eevees is transfering all of them even when they are over the 80% iv and i have not to transfer over 80% iv. I already lost a flareon, vaporeon, jynx and many eevees for these. Doesnt matter at least I got my dragonites :) 
thanks for the good job and nice bot Im just adding info to check in future updates

bornsupercharged:
Redundant with PR1495, additional issues are resolved in that one :)

PR #1492, opened 2016-07-30
Adding "both" as an option in LevelUpByCPorIv

Artearte:
I added it to LevelUpPokemonTask but it can also be done in the DisplayPokemonStatsTask, I had to add collections and linq directives so that might not be graceful


Artearte:
Uh, this actually breaks, I probably did something very stupid, I'll look into it later as now I have to go

PR #1494, opened 2016-07-30
Impltement optimized recycling

BadWolfL:
If the amount of total balls currently hold is greater than
TotalAmountOfPokebalsToKeep, the Bot will recycle poke balls from low
level balls.

In Configs, TotalAmountOfPokebalsToKeep should be less than the sum of
all separate poke balls in ItemRecycleFilter.

Now optimized recycling of potions and revives also works in latest commit.


PoGo.NecroBot.Logic/Tasks/RecycleItemsTask.cs
```diff
@@ -29,7 +30,268 @@ public class RecycleItemsTask
                 DelayingUtils.Delay(session.LogicSettings.DelayBetweenPlayerActions, 500);
             }
 
+            if (session.LogicSettings.TotalAmountOfPokebalsToKeep != 0)
```
navidemad: TotalAmountOfPokebalsToKeep
by
TotalAmountOfPokeballsToKeep
👍

PR #1495, opened 2016-07-30
Fix bugs with ball logic to use proper pokeball type

bornsupercharged:
There were numerous issues with the ball logic code, which have now been resolved. The configuration file also has pokeball-type catch probabilities now, so if you only want to use great balls on 20% and ultras on 10%, you can do that. Defaults are set to as they were hardcoded previously.


DurtyFree:
Some of these  bugs got already fixed, please merge or resubmit the pr.

bornsupercharged:
This is ready to go now.

PR #1498, opened 2016-07-30
Refactor upgrading

bmdr38:
Changed upgrading to only upgrade when enough candies are available and only when you have enough spare candies to evolve either the pokemon in question:
Example: gloom can still evolve to vileplume, so dont upgrade
or for your others in your inventory:
Example: you have flareon/vaporeon etc, but also have a really high iv/cp eevee that needs evolving, it will keep candy to evolve the eevee before upgrading the rest
Note: will still only upgrade once per 5 stops, but will instead do all eligible once.  It will eventually max out upgrading over time

Removed garbage from displayPokemonStatsTask

More frequently display pokemon stats to track overall progress instead of only at start

Added stardust and level checks as well to avoid trying to needlessly upgrade a pokemon when you have no stardust or it is at "max" (player level + 2)

### DIFF
--- a/PoGo.NecroBot.Logic/Inventory.cs
+++ b/PoGo.NecroBot.Logic/Inventory.cs
@@ -354,6 +354,12 @@ namespace PoGo.NecroBot.Logic
                     .Where(p => p != null && p.FamilyId != PokemonFamilyId.FamilyUnset);
         }
 
+        public async Task<IEnumerable<PokemonUpgradeSettings>> GetPokemonUpgradeSettings()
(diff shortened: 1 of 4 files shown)


tehaon:
ahhh this is definitely going to conflict with the pull request I have in.

bmdr38:
Im fixing the stardust problem right now as well, but yea.  Ill leave it up to the admins on which they want to take.  I have no preference.

tehaon:
This is now merged into my pull request.

PR #1499, opened 2016-07-30
Fix custom pokeball

AlessioDP:
Without these custom pokeball on capture will works.

Right now every pokémon with 300 cp or more will use a great (or better) ball, bypassing custom pokeball settings.


AlessioDP:
If someone wanna improve percentage system on capture is better!

DurtyFree:
Already got fixed.

PR #1505, opened 2016-07-31
Write location to disk when updating so we can read it on start and p…

jessewhite007:
prevent a soft ban when you change your default location to something too far away from your last position

Update directory we look in for the coordinates from configs to config since everything else is in config


NecronomiconCoding:
please reopen because plebs filled this with crap

PR #1506, opened 2016-07-31
Added logic to wait for recycling until the inventory reached a treshhold.

th3ragex:
Hi,

i have changed the source code to wait until the inventory fills up to a configureable treshhold before recycling. I think this is more similar to a human player and reduces the recycling because the bot does not have to recycle all ~5 pokestops.

I have choosen 20 as default treshhold, it worked for me, maybe it needs some recalibration after some user feedback.


tangdami:
Guys, if you got "Account not verified error" with PTC account, because the lastest release swapped the lines of username and password.... It took me hours to figure it out..

PR #1511, opened 2016-07-31
DPS 

alexdilu:
Calculate DPS per move. Need to rewrite decently and update missing moves.
Source: https://thesilphroad.com/research


AdamWeishaupt1776:
not sure if its your indent settings, but these changes should not be affecting the entire file.

NecronomiconCoding:
^

PR #1515, opened 2016-07-31
Minor code cleanup

rfaes:
Replaced Console.Writeline with Logger.Write and Thread.Sleep with Task.Delay


NecronomiconCoding:
fix conflicts :smile:

PR #1530, opened 2016-07-31
Update FarmPokestopsTask.cs

wbayer81:
changed inventory cleanup after fixed 5 pokestops to random value between 3 and 9 to make it harder to detect


Epinikion:
compiles great and is a good "flying under the radar feature"

PR #1535, opened 2016-07-31
Use Lucky Egg and incense constantly

cawk:
Option to use luckyegg and incense constantly wont waste them either so will renew every 30 minutes can be disabled if wanted


NecronomiconCoding:
Logger > Translation pls

cawk:
going away for a week will fix the translations then 
if anyone wants to use this code to add the translations go for it if not ill do it when i get back

PR #1536, opened 2016-07-31
Fix a syntax issue

yvan-sraka:


Aneduna:
I am Sorry for that one, i failed it because i was frustrated with github. Next time i will care more.

PR #1550, opened 2016-07-31
Add InnoSetup installer

narfman0:
### Description
In addition to Release.zip, perhaps we could have an installer release to add start icons, uninstall, and related. Attached is a trivial inno setup (common windows foss installer) install script. I don't know how release.zip is made, but this script could be run on that result.

### Testing
1. Compile script from 4.0 extracted Release.zip
2. Execute shortcut from start menu
Verified Logs and temp did not copy over. Ensured app ran (very briefly).

### Note
Pardon I don't know more about process of how release.zip is made, but I'm all ears. Was crazy fast for me to write it but I can make tweaks too! :)

### DIFF
--- a/installer.iss
+++ b/installer.iss
@@ -1,0 +1,19 @@
+[Setup]
+AppName=NecroBot
+AppVersion=1.0
+DefaultDirName={pf}\NecroBot
+DefaultGroupName=NecroBot
+UninstallDisplayIcon={app}\NecroBot.exe
+Compression=lzma2
+SolidCompression=yes
+OutputDir=.
+OutputBaseFilename=NecroBotSetup
+
+[Files]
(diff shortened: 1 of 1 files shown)


A-Charvin:
Does the bot really require an Inno setup wizard? The same function can be achieved by simply using an SFX archive mode. It's simple to make an Installer with Inno though. :+1: 
I can help with that if needed though.

OuttaControlX:
sounds excessive tbh.

VanStorm:
I think there is no need for creating an installer- Like @A-Charvin already pointed out, an SFX archive would give us the same experience but imho it works fine as it is.

Simpuhl:
Don't think we need an installer either, but if you want include it as optional part of the release that would be cool. 

Might confuse people on where to find the conf and auth though

Platem:
Maybe adding a shortcut to config in the app start folder would fix that, Simpuhl.

NecronomiconCoding:
Thanks for the work, but we have decided this is too overkill.

PR #1555, opened 2016-07-31
Some minor fixes and improvements

actislav:
Some minor fixes and improvements


actislav:
EvolveKeptPokemonsAtStorageUsagePercentage config parameter added, only works in tandem with KeepPokemonsThatCanEvolve (with value 0 everything will be the same)

CalculateMaxCp now are using just PokemonId instead PokemonData

Serialization config settings added and changed behavior to persist even default values of object

Snipe in area will use sorted pokemons by MaxCP (maybe, first gather all data from all areas and then sort, for be sure not to use your balls on crap)

Prevent use incorrect player position after possible exception occurred in snipe logic by next state of StateMachine

PR #1559, opened 2016-07-31
Events: Add info to eggslist event to display accurate data

nicoschmitt:


TheTravelingTrainer:
Shouldn't this be sent with the profile data? Or am i missing something?

nicoschmitt:
The GUI can ask for the current list, this is what this event is for

PR #1586, opened 2016-07-31
LevelUpPokemonTask

keinPlan:
only check pokemonlist that is used to upgrade (iv/cp)


NecronomiconCoding:
a better description please. @keinPlan

PR #1587, opened 2016-07-31
Write location to disk on update

jessewhite007:
prevent soft bans when default location is too far away from last location
change read location of coordinates to config where everything else lives


NecronomiconCoding:
can you please rename from Coords.ini to LastPos.ini

jessewhite007:
Done

actislav:
maybe i am wrong, but that code does not prevent, just warn

jessewhite007:
Good point, I could either throw an exception or just set to last position.

Rawrlex:
Working with .gpx too?

jessewhite007:
@Rawrlex yes it works with .gpx too.

PR #1598, opened 2016-07-31
Fully customizable item recycler

recon88:
Thought we should bring back the **fully customizable** recycler instead of having max combined potions, revives, etc. So you can set how much of every item you want to keep. Makes more sense than having potions and revives combined, because some people only would like to keep max revives/potions and use them at gyms after botting. Saw some asking for that on the discord.


recon88:
Nevermind. Will make another PR which allows to switch between recycle behaviours

PR #1609, opened 2016-07-31
Update Translations for egg and incense

cawk:
Adds translation for new tasks
done via github website and on phone 

EDIT works


Oolaruu:
Clearly not sure about
GetPokeDexCount.cs
"AmountPkmSeenCaught"
"Amount of Pokemon Seen:{}:151, Amount of Pokemon Caught:{}:151"

I'm pretty sure there are a number inside {} but which one.
I made my translation like this.
"Pokemon Deck(151) : Seen {} | Caught {}"
to have less "Amount of Pokemon" + ":151" twice

simplyphp:
Fix the conflicts and we can re-evaluate.

PR #1619, opened 2016-08-01
Improve for using GreatBalls and Ultraballs

ahmadsafar:


lolcheat321:
I appreciate the logic of what is being attempted, but wouldnt it be better to leave the OR in and simply AND a check that the minIV isnt 0 to the minIV logic? That would solve the current config file settings error. Whereas, your current change will never trigger on catch probability unless it also satisfies the minIV.

PR #1639, opened 2016-08-01
Change CSV dump to be comma seperated-values as well as output more data

Dgc2002:
The first row is the header for the columns  
There are some properties of ```PokemonData``` that I left out since they didn't seem applicable.  
The information currently dumped isn't very useful mostly due to the way it's formatted.


salvob41:
issue #1687

PR #1646, opened 2016-08-01
Clean up PokeDex to print out correct findings.

dddbliss:
Small refactor of pokedex code to show correct numbers, and also changes it from using string manipulation to find pokedex entries.


ChronoXNL:
This pull request also solves a potential "Index was outside the bounds of the array." issue.

PR #1664, opened 2016-08-01
Snipefix

OuttaControlX:
pokemon box gets full when you use snipe to power farm then just spams
inv full rather than initiate evolve>transfer, forcing you to restart
the bot to do it.


OuttaControlX:
totally efed up my pull request.

NecronomiconCoding:
So what exactly was changed here?

chandler14362:
All you did was add and remove spaces, how does this fix anything?

PR #1667, opened 2016-08-01
Inventory Full During Snipe Power Farming Fix(for real this time)

OuttaControlX:
pokemon box gets full when you use snipe to power farm then just spams
inv full rather than initiate evolve>transfer, forcing you to restart
the bot to do it.


erikeleria:
Someone merge this. This is a huge fix for the full pokemon inv while sniping.

PR #1714, opened 2016-08-01
GetBestBall tweaks

l3uddz:
Added ThrowLuckyBallFirst, this will always throw a pokeball on the first catch attempt, if the pokemon matches ultraball criteria, it will throw a greatball.

Added StrictPokeballUse, this stops throwing pokeballs that dont match the specified criteria (IV/CP).

Added PokemonToUsePokeball list, when StrictPokeballUse is false, it will only use a pokeball on these types of pokemon.

### DIFF
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -140,9 +140,13 @@ namespace PoGo.NecroBot.Logic
 
         ICollection<PokemonId> PokemonToUseMasterball { get; }
 
+        ICollection<PokemonId> PokemonToUsePokeball { get; }
+
         Dictionary<PokemonId, TransferFilter> PokemonsTransferFilter { get; }
         SnipeSettings PokemonToSnipe { get; }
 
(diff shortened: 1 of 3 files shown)


Aneduna:
This is Problematic with Pokemons you really want, for example 100% vi Pokemons
You may catch them with an ultra ball
but if you throw a pokeball, you lose the 100%.

The Option itself is nice tho

l3uddz:
If its turned off (which it will be by default without specifically turning it on) it will continue as normal.
If the pokemon matches ultrabrall cp/iv criteria, it will throw a greatball first instead of a pokeball.

It actually works rather well, my ultraball settings is for 2000 cp or higher, 99% iv or higher and im always seeing things like:

[11:06:58] (PKMN) (CatchSuccess) | (Normal) Snorlax Lvl: 29 CP: (2529/2579) IV: 86.00% | Chance: 11.12% | 935239.82m dist | with a GreatBall (49 left). | Candies: 1632

with it turned on, typically because of that CP it would have used an ultraball, whereas evidently sometimes its not required.

simplyphp:
Fix the conflicts and we can re-evaluate.

PR #1731, opened 2016-08-01
Pull request for refactoring of the logging messages

igeligel:
**What i have done:**

- Created Model for logging strings which can be initialized by a static method
- In the method use null propagation together with ternary operator to set the default string or the translated string
- Changed access modifiers of the constructor of the logger to internal
- Fixed indentation of xml documentation comments and normal comments to fit [official guidelines](https://msdn.microsoft.com/en-us/library/ff926074.aspx#Anchor_2)
- Added new line at end of the file
- Update Access Modifiers of the Console Event listener and change the parameter names there

### DIFF
--- a/PoGo.NecroBot.CLI/ConsoleEventListener.cs
+++ b/PoGo.NecroBot.CLI/ConsoleEventListener.cs
@@ -1,6 +1,7 @@
 #region using directives
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using PoGo.NecroBot.Logic.Common;
 using PoGo.NecroBot.Logic.Event;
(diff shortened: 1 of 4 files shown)


Aneduna:
Great job. Did anything in the output change or just in the usage/internals?

igeligel:
There were no changes. I have make a [picture](http://i.imgur.com/1vj8RlS.png) of the current console output.

Output strings were not changed but i was close to fix a typo for the transferred string but i do not know any guideline on those strings (like max length).

```csharp
Transferred =
                session?.Translation.GetTranslation(
                    TranslationString.LogEntryTransfered) ?? "TRANSFERED";
```

Should be written with two r's i believe. If it is allowed ill change it too.

PR #1735, opened 2016-08-01
CLi Size

toquengtoketin:
temporarily and easily expand the cli for readibility?


TankTheFrank:
The console window properties can be set as part of the OS, along with font size, colours, etc. Overriding a configurable OS property with a hardcoded value is not a good idea.

TankTheFrank:
Just to be sure that you understand what I am talking about - click on the icon on the top left of the console window, properties/layout. The window size set there is preserved between runs - or used to be/should be, now is resized at every exe launch.

PR #1754, opened 2016-08-01
Pokestop Time Until Arrival

Andrerm124:
Implements a feature to display how many seconds it should take to
arrive at a pokestop.


SupaCURE:
Haha it would be especially useful for all of us who just like watching it work, there's something relaxing watching a program run some things in the background.

NecronomiconCoding:

PR #1772, opened 2016-08-01
Updating IV and CP logic for transferring.

dddbliss:
When choosing Pokemon to transfer, if IV, then keep highest CP - if CP, keep highest IV.


Silverarrows1986:
come on i need this :(

Silverarrows1986:
or something to choose  

"PrioritizeIvOverCp": true,
"PrioritizeAfterIvByCpOrStamina": "cp", 

dunno if its possible to do

dddbliss:
Relax, the PR is still open.

Algent:
Awesome, is there any way you could also (In another PR maybe ?) add an option to keep both best IV and best CP unless it the same pokemon ?

Issue #1753, opened 2016-08-01
[FeatureRequest] Transfer Pokemon based on MOVES


yoshi2323:
I need to get rid of high IV pokemon with the wrong moveset, my accounts are bursting....
Best thing for me would be a list to add pokemon + their desired moves + individual IV cutoff:

"Dragonite": { "Move1": "DragonBreathFast", "Move2":"DragonClaw", "minIV":95},
........

edit: maybe we just need to add move1 and move2 attributes to the "PokemonsTransferFilter" to achieve this

yoshi2323:
my idea was to check for move1 AND move2, to filter out pokemon where BOTH moves are best

burgerbutts:
As of right now it seems to be keeping pokemon of any CP/IV value as long as they have a single skill from the move list.

There should definitely be a way to consider which pokemon the bot should keep, and an option to consider one or both moves. Something similar to this, I would assume:

- check settings to use single or both moves
- has single or both moves
- check PrioritizeIvOverCp
- compare to find the best
- transfer the rest

PR #1807, opened 2016-08-01
Add Moves to TransferFilter

sawirth:
Based on the idea of #1753 I added a new property "Moves" to the TransferFilter. If a duplicate Pokemon has one of these moves it won't be transferred.

If you accept this PR, the wiki should mention that the first move always has the postfix "Fast". Otherwise it can't be deserialized.


yoshi2323:
my idea was to check for move1 AND move2, to filter out pokemon where BOTH moves are best

sawirth:
the logic atm is that it makes an OR check for all filter properties so I didn't want to change that only extend the current behavior. With this behavior there is no need for two separate properties.

simplyphp:
Merge conflict

Pauloozzy:
how use this movies filter?

PR #1845, opened 2016-08-01
Corrected Spelling

vanillasteel:


simplyphp:
This is used in more than just those files, incomplete as-is.

PR #1852, opened 2016-08-01
Snipe task: prevent incorrect player position setting; Evolve task, fixed never evolve case; Null reference exception in SnipeScanForPokemon fix

actislav:
splited getpokedexcount fix from last pull request


actislav:
Also Null reference exception in SnipeScanForPokemon ScanResult check fix

actislav:
Fixed never evolve potential case, fixed logic of evolve

PR #1860, opened 2016-08-01
Fix Pokemon never evolving

sollematthew:
Pokemon now evolve if KeepPokemonThatCanEvolve is set properly.


actislav:
remove this pull request, its wrong code, check my instead

PR #1874, opened 2016-08-01
[Feature] Added a better config merging system

Andrerm124:
Fixed conflicts and reimplemented:
Ignore all pokemon not on Sniper Filter
Time for pokestop arrival

### DIFF
--- a/PoGo.NecroBot.CLI/ConsoleEventListener.cs
+++ b/PoGo.NecroBot.CLI/ConsoleEventListener.cs
@@ -113,9 +113,11 @@ namespace PoGo.NecroBot.CLI
 
         private static void HandleEvent(FortTargetEvent fortTargetEvent, ISession session)
         {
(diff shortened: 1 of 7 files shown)


Andrerm124:
I'd really like to have a second opinion on whether my method should be the new standard and whether I should rip out the old implementation.

dddbliss:
If you look in Settings.cs in Logic, around line 482 you will see where we apply a bunch of defaults, if you can verify we no longer need these and can use DefaultValue attributes insteads, that would be awesome.

Andrerm124:
@dddbliss I did what you suggested and it seems to work fine.

I think this method may overwrite a lot of the VersionCheckState code (Specifically TransferConfig())

Andrerm124:
Do not commit, found an issue with fresh installs. Trying to resolve issue.

simplyphp:
Let us know when the issue has been resolved.

Andrerm124:
Already fixed, it's ready for a commit I believe. Unless you can see any issues that is.

Andrerm124:
@simplyphp Could you please commit this request? I'm heading off for the night and won't be able to fix any conflicts that pop up, thanks :)

Issue #1801, opened 2016-08-01
Configuration Migration System issues (upgrading from 0.4.0 to 0.5.0)


Jertlok:
Hello there, i've noticed that all the new configurable settings (included in 0.5.0) get a 0 parameter as soon as the settings migrate during the bot automatic update from version 0.4.0.
_Example:_ 
```
  "UseGreatBallBelowCatchProbability": 0.0,
  "UseUltraBallBelowCatchProbability": 0.0,
  "UseMasterBallBelowCatchProbability": 0.0,
```

## Information

**Bot version:** NecroBot 0.5.0 taken from _Releases_
**Operating System:** Windows 

## Steps to Reproduce

-  Opened NecroBot version 0.4.0
-  Waited for the update procedure and config.json migration:
```
[17:36:53] Downloading and apply Update...
[17:36:59] Finished downloading newest Release...
[17:37:00] Finished unpacking files...
[17:37:00] Update finished, you can close this window now.
[17:37:00] Finished transferring your config to the new version...
```
- Checked the new generated config.json file - _and the faulty config.json was there_

sniper85500:
Bot always use Greatballs now - has anyone the original Values from the last version?

Jertlok:
```
  "UseGreatBallBelowCatchProbability": 0.5,
  "UseUltraBallBelowCatchProbability": 0.4,
  "UseMasterBallBelowCatchProbability": 0.05,
```

sniper85500:
It Works - thank you very much :-)

PR #1879, opened 2016-08-01
Fix default ball settings.

dddbliss:
Fixes for default pokeball settings for #1801


Linewalker:
You should use switch in case of if [] for the balls ... better to read.

PR #1882, opened 2016-08-01
Set Console.Title on start

j2ghz:
When it's waiting a long time to login, it just shows path to the exe, now it'll show NecroBot starting.


AiksuCoding:
Fix conflicts :smile:

PR #1906, opened 2016-08-01
Update .gitmodules to use directly original FeroxRev api and not forks

hilycker:
Update .gitmodules to use directly original FeroxRev api and not forks


simplyphp:
This will not work because FeroxRev API does not have the changes yamashi wrote.

simplyphp:
Just saw that they actually did merge it in, re-opening.

simplyphp:
[DO NOT MERGE THIS YET]

dddbliss:
Merging into our own for a while until Ferox can fix csproj related issues.

PR #1946, opened 2016-08-02
Experimented with replacing own logging with NLog

RoseFlunder:
This is just a proof of concept if the current handcrafted logging could also be done with NLog logging framework.
I did my best to rewrite as less as possible and simplifying setting up the logging.
What do you think? Is it better or worse than before?
I like that an own ConsoleLogger is not needed anymore and File logging is also done by NLog, but the downside is that the colors for console logger are tied in the NLog Config file now. I dont know if it is possible to overwrite them in the code on demand.
I am no expert at either c# nor NLog, it is just an interesting experiment for me and I wanted to share the results and get some feedback from you.

EDIT: I moved the Logging class from Logic to CLI. Logic always sends events and the CLI project is responsible what to do with events and may log them.


RoseFlunder:
EDIT2:
Second thought: Maybe not as easy because the Logger class itself is still in the Logic project and i think it should not know at all that there is sth. like a ColorConsoleAppender because the targets are defined by the CLI project. So just simple logger calls are all what should "allowed" be allowed in the logic project.

Maybe it would be even nicer if the whole Logger class would be in the CLI? But how would logic log then? own logging file for errors, without translation or sth. via NLog loggers, just for dev purpises and everything the user should see via events insteads like to call Logger.write(...) directly? What do you think would be a nice and clean approach?

Magnog:
Is this available on mac?

simplyphp:

PR #1947, opened 2016-08-02
Config rewrite

davegranlet:
Moved options around to better group together like options.
Still needs formatting. ( ie section heads like ---General etc )


simplyphp:
Sorry but this is totally outdated as compared to the current Settings.cs, would highly recommend you look at what we have as it stands right now, thank you

PR #1979, opened 2016-08-02
Fixing the submodule with latest from FeroxRev.

dddbliss:
Fixing the submodule with latest from FeroxRev.

Also corrected some ugly merging issues.


dddbliss:
Hold on this until I can add exception tracking for LoginInvalidException that the API now throws.

hilycker:
Could we link directly to FeroxRev repo instead of a fork?

dddbliss:
No, due to the fact that FeroxRev's sln does not compile currently due to broken references in his solution files.

PR #2031, opened 2016-08-02
Add new option 'AlwaysEvolveWithLuckyEggWhenMinPokemonAmountOccurs' w…

redChrillz:
I wanted a way to use the nice logic for KeepPokemonsThatCanEvolve = true and use a lucky egg when the UseLuckyEggsMinPokemonAmount criteria was met.

I therefore introduced a variable called 'AlwaysEvolveWithLuckyEggWhenMinPokemonAmountOccurs' which can be used together with KeepPokemonsThatCanEvolve = true.

We should maybe discuss what we want to do here if you dont like my idea. I think that KeepPokemonsThatCanEvolve  is good but i wanted to use it to _not_ transfer pokemonts and only use a lucky egg when UseLuckyEggsMinPokemonAmount  was met.

Any input?


simplyphp:
Please pull in master and fix the conflict, then resubmit the PR

PR #2016, opened 2016-08-02
Fixed typos + French

Oolaruu:
Typos:
itemsToRecylce -> itemsToRecycle
TotalAmountOfPokebalsToKeep -->TotalAmountOfPokeballsToKeep

French pre-updates:
eventFortUsed #2037 
eventPokemonCapture #2035

PR #2037, opened 2016-08-02
Add lat and long to EventFortUsed event

redChrillz:
Added lat and long info to pokestop output so its easier to follow the bot and know where it is


Oolaruu:
#2016

PR #2053, opened 2016-08-02
Update SnipePokemonTask.cs

zzzkl:
fix can not snipe bug


recon88:
I suggest something like this instead of just that short Header:
"Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2228.0 Safari/537.36";

recon88:
Also, this has been already fixed.

PR #2054, opened 2016-08-02
Add always evolve with lucky egg when min pokemon amount occurs option

redChrillz:
Follow up on https://github.com/NECROBOTIO/NecroBot/pull/2031.

Fixed merge conflict


redChrillz:
What are the guidlines for documentation?

I.e. i have added a new flag here 'AlwaysEvolveWithLuckyEggWhenMinPokemonAmountOccurs' and it should have some documentation

simplyphp:
Has this been thoroughly tested? Right now the lucky egg functionality works fine, with the maximum amount of XP gained. What you've implemented works fine for people who are not trying to maximize their XP gains, however most people are, so I would at a very minimum have your default setting as false.

redChrillz:
I was looking for a discussion on the subject :)

I wanted the possibility to evolve using lucky-eggs when we hit the value in 'UseLuckyEggsMinPokemonAmount' and having 'KeepPokemonsThatCanEvolve' set to true.

In the current implementation if both are set we will have to wait until the value EvolveKeptPokemonsAtStorageUsagePercentage is hit.

So yes we do lose XP since we will not fill the pokemon-slots but we will evolve with regards to the 'UseLuckyEggsMinPokemonAmount' parameter.

But maybe this is a bad idea and we should only use EvolveKeptPokemonsAtStorageUsagePercentage and basically fill up with pokemons before we evolve to maximize the egg usage?

I started with this issue before EvolveKeptPokemonsAtStorageUsagePercentage existed.

simplyphp:
Ah okay, yeah I do like your implementation of a function to handle the egg usage logic though, maybe you can retain that but take out the AlwaysEvolveWithLuckyEggWhenMinPokemonAmountOccurs?

redChrillz:
Sure, so the logic would be,
KeepPokemonsThatCanEvolve = true will look only at EvolveKeptPokemonsAtStorageUsagePercentage.

That should mean that we transfer pokemons when we cannot evolve them and do a mass-evolve when the value specified for EvolveKeptPokemonsAtStorageUsagePercentage is hit.

Sounds good to me

redChrillz:
See the new commit and diff

PR #2091, opened 2016-08-02
RE-organized the config order to make it easier to find options

davegranlet:
RE-organized the config order to make it easier to find options
Still looking for a usable option to add section breaks


simplyphp:
Necro is going to do settings differently, please halt further work on this as it likely wont be accepted.

PR #2100, opened 2016-08-02
Added Level Up and Item Rewards logging

Andrerm124:
Every time you level up, it is displayed as a message, and also if there are any items rewarded, they will be listed.

Somebody could do me a favour and set up a config option for this; I don't have the time for that right now.


Andrerm124:
Just to clarify, this is built upon @cawk's work and a lot of the credit goes to him

cawk:
Nice work I completely forgot about this was meant to work on it but good work :D

PR #2126, opened 2016-08-02
Turning weak pokemon transfer off by default

simplyphp:


RinneganMadara:
Why does/did weak transfer override KeepMinDuplicatePokemon? I just had my bot transfer 80+ pokemon.

PR #2213, opened 2016-08-02
prevent wrong Evolve behavior, pls do n't change meaning of mnemonic parameters

actislav:
pls, dont break the logic of evolve behavior, analyze and test your changes before commit to master, Evolve*KEPT*PokemonsAtStorageUsagePercentage have only one certain meaning. If you want change logic then pls rename parameter to EvolvePokemonsAtStorageUsagePercentage without KEPT.

Without this check people who wants to evolve pokemon with above IV or just with enough candy, but with options KeepPokemonsThatCanEvolve = false (case with default config for new users) will never have it, and even their pokemon might be transfered instead of evolving... (sry for my bad english)


BennettStaley:
Fantastic work. Working as it did on ver 5.0.

PR #2219, opened 2016-08-02
wrong KeepMinOperator logic in Inventory.GetDuplicatePokemonToTransfer

s-p-c:


simplyphp:
My original logic is correct:
If a pokemon has 2000 CP and 99 IV, but we only want to keep those that are 2k and 100 IV, then we dont want to do 2000 < 2000 && 99 < 100 because it wont be met

simplyphp:
If you're trying to get a list of pokemon that should be transferred, and you ONLY want to keep pokemon that are 2000  AND  100IV, then checking that cp is < 2000 AND that iv < 100 you will miss out on every pokemon where cp is < 2000 but iv is100 - meaning you're going to keep a 100iv pokemon with 100cp, which is not what the settings are.

s-p-c:
I'm very sorry, you are right of course!

PR #2237, opened 2016-08-02
Version should be 0.6.1 i guess

somani6:


SweetCherry:
**# Yes this should be fixed after v0.6.1, so be patience and await the next update....

Thank you!**

PR #2367, opened 2016-08-03
Inventory transfer function refactoring, improve and fixes of some wrong behaviors

actislav:
transfer pokemon task refactor, faster and better code, main feature counting pokemon that we can keep for evolve according to new candy count from some of pokemon that we will only transfer at this moment


simplyphp:
Why did you leave out their movesets?
pokemonTransferFilter.Moves.Contains(duplicatePokemon.Move1))
pokemonTransferFilter.Moves.Contains(duplicatePokemon.Move2))

Please review the code and see what was removed, and factor it back in.

PR #2446, opened 2016-08-03
fix transfer bug ( usage of global KeepMinOperator instead of local)

actislav:


Naprosnia:
I ran this change and it seems to be working finally.  ;)

Issue #2140, opened 2016-08-02
Please add random speed (ex: 15-40)


aliceaia:
Please add random speed (ex: 15-40). This number will automatically change each time moving to pokestop. Thanks so much.

Sylencer90:
Requested this a while ago, like on 0.3 or sth. Has never been added tho.

aliceaia:
I think this will be better than we move to a fixed rate.

RoseFlunder:
How would you specifiy it? Two settings like:
MinWalkingSpeedInKilometerPerHour
MaxWalkingSpeedInKilometerPerHour
Or do you have another idea?

aliceaia:
@RoseFlunder
Exactly. And the tool will randomly select a number between min - max for each move.

ANR2ME:
Slightly randomize the Altitude would also be nice, together with speed.

RoseFlunder:
Working on
MinWalkingSpeedInKilometerPerHour (default value?)
MaxWalkingSpeedInKilometerPerHour (default value?)

PR #2460, opened 2016-08-03
added support for random walking speed

RoseFlunder:
PR is related to this issue: [https://github.com/NECROBOTIO/NecroBot/issues/2140](https://github.com/NECROBOTIO/NecroBot/issues/2140)

two new settings:
"MinWalkingSpeedInKilometerPerHour": 31.0,
"MaxWalkingSpeedInKilometerPerHour": 31.0,

per default both settings are set to the same value so its just as the old behaviour -> constant speed
but one can vary the values now to achieve a random walking speed.

One thing that would be nice is missing yet:
Transfer the old "WalkingSpeedInKilometerPerHour" of the users to the new settings. It just sets the defaults of 31 now, which is also the default WalkingSpeedInKilometerPerHour if nothing is set before.

If you can answer me how to do it, i can make another commit.

### DIFF
--- a/PoGo.NecroBot.CLI/Config/Translations/translation.de.json
+++ b/PoGo.NecroBot.CLI/Config/Translations/translation.de.json
@@ -39,7 +39,7 @@
     },
     {
       "Key": "eventFortTargeted",
-      "Value": "{0} in ({1}m)"
+      "Value": "{0} in ({1}m)(~{2} Sekunden)"
     },
     {
       "Key": "eventProfileLogin",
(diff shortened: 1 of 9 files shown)


Oolaruu:
#2140

MegaThorx:
https://github.com/NECROBOTIO/NecroBot/issues/2140 correct link to the issue

Sdwz:
For god sake, use an offset instead of min max, and speed +  random from -offset to offset.

ChemiHatake:
what's the range for the random?

PR #2516, opened 2016-08-03
Update translation.es.json

antolinjm:


javilonas:
La frase original dice: "Amount of Pokemon Seen: {0}/151, Amount of Pokemon Caught: {1}/151"

No dice: "Number of Pokemon Seen: {0}/151, Number of Pokemon Caught: {1}/151"

Issue #1895, opened 2016-08-01
[Feature request] Customizable throws settings


GediminasMasaitis:
Inspired by #1872 , but not quite what was suggested there.

I'd **love** fully customizable throw settings! For example:

```
MissThrowChance: 0.15,
NiceThrowChance: 0.25,
GreatThrowChance: 0.15,
ExcellentThrowChance: 0.02,
CurveThrowChance: 0.10,
ForceGreatThrowAboveIV: 95.0,
ForceExcellentThrowAboveIV: 99.9,
ForceGreatThrowAboveCP: 1800.0,
ForceExcellentThrowAboveCP: 2500.0,
```

Maybe even settings for individual Pokemon - although that may or may not be overkill.

Just need to be careful with forced throws - if capturing a Pokemon with a forced throw and a very low capture chance, which keeps escaping - you may end up doing 10 excellent throws in a row - which would be a clear indication of botting. Maybe only allow forced throws on the first attempt or something.

PR #2536, opened 2016-08-03
Humanized throws

Codecamv:
Added customizable throws as current version throws excellent curve balls 100% time. New config. values to adjust the rate. Currently all throws hit but can be distributed through settings.

#1895 

I'm not sure how to remove the translation file from the change set.


simplyphp:

PR #2548, opened 2016-08-03
Customizable throws

Codecamv:
Added customizable throws as current version throws excellent curve balls 100% time. New config. values to adjust the rate. Currently all throws hit but can be distributed through settings.

#1895


Codecamv:
Sorry about that, I'll try to explain the commit better.

This commit introduces more human-like pokeball throws. If you enable HumanizedThrows you may also specify the percentage of different quality throws; nice, great or excellent hits (bonus). You can specify the percentage of throws that are curved (spin bonus) as well. 

By turning on humanized throws you will use more pokeballs, also more pokemons will flee, because bonuses affect catch rate. However, throwing excellent spin balls 100% of the time is very bot-like and may be flagged by Niantic. 

Config explanation (hopefully soon in wiki):
EnableHumanizedThrows (true/false) 
'false' is default; all balls will be thrown at Pokemons and hit with excellent bonus and curve/spin bonus (just like before this commit).
'true' enables humanized throws where bot throws anything from regular to excellent hits.
  
NiceThrowChance
Percentage of nice hits

GreatThrowChance
Percentage of great hits

ExcellentThrowChance
Percentage of excellent hits

CurveThrowChance
Percentage of curve/spin balls

ForceGreatThrowOverIv
Forces a great throw if IV% is over specified value

ForceExcellentThrowOverIv
Forces an excellent throw if IV% is over specified value

ForceGreatThrowOverCp
Forces a great throw if CP is over specified value

ForceExcellentThrowOverCp
Forces an excellent throw if CP is over specified value


Hope this makes more sense @salvob41 ?

Issue #2335, opened 2016-08-03
[Feature Request] New user, filled up Auth.json and few config.json in console


Oolaruu:
Hello,

Another request that I can suggest specially for new user, who have just downloaded the bot (not from an update) would be:

- Welcome message
- Choose your AuthType:   [1] Google or  [2] Ptc
- Write your Username
- Write your Password
- Choose your destination: [1] Co-ords1 or [2] Co-ords2 etc
Specially for based region pokemon
- Confirm

Or something similar.

Save Auth.json + Config.json, Autoclose, Restart
Enjoy

Thanks

Andrerm124:
I'd assign myself to this issue if I could. I'm almost completed and will reference this issue when pushing.

PR #2566, opened 2016-08-03
[Feature] Added config initialisation & Resolved Issue #2335 

Andrerm124:
This request adds a system to allow the user to generate a custom config file as suggested from issue report #2335.

This system gives the user the option to skip this process however.


Oolaruu:
A big thank you. I do and others with me will appreciate for sure.

Andrerm124:
@Aneduna The final commit I made only removed the `"This is your first start and the bot has generated the default config!" line because I had implemented the (almost exact) same message elsewhere (When the user chooses to skip the config builder to be exact)`

PR #2647, opened 2016-08-03
added Setting 'DoNotTransferFavorite' / German Translation for 'WaitingForMorePokemonToEvolve'

somani6:
added Setting 'DoNotTransferFavorite' with default value 'true' for customizing if favorited pokemons should be considered for weak-transfer and duplicate-transfer or not.

added missing german translation for 'WaitingForMorePokemonToEvolve'


simplyphp:

PR #2677, opened 2016-08-03
Update translation.nl.json

CooOooL74:


StevasaurousREX:
Add the first quote back on line 34. It is causing errors.

simplyphp:
Code fails to compile.

PR #2708, opened 2016-08-04
Delay between evolves

ghost:
Added delays between evolves to humanize the process.


Pym:
Isn't this delay very short..? I think the duration of an evolution is about 20 seconds (because of the animation). If we want to really look human, it might be better to set a bigger delay between evolutions.

PR #2709, opened 2016-08-04
Adding exception handling

simplyphp:


poutros:
@BenMeitav You know how there is text below all those files in [here](https://github.com/NECROBOTIO/NecroBot). Read that text.

PR #2714, opened 2016-08-04
added move operator in transfer filter, fixed keepminlevel logic, added some comments

actislav:


actislav:
pls dont close PR in conflicts case, mark client action needed

simplyphp:
It's probably safe to assume this code is untested, given the current situation?

actislav:
tested because created 1.5 day ago

PR #2720, opened 2016-08-05
Update translation.nl.json

karoi:
Someone changed 'Snoepjes' in to 'Candies', so I changed the other 'snoepjes' in to 'candies'. So it's consistent.


rickerd:
Candies is better yes. "Snoepjes" is a little bit to litterally translated

PR #2734, opened 2016-08-07
Update translation.fr.json

SebastienBenazet:


Oolaruu:
Thank you. as always good translation.

SebastienBenazet:
I was looking to mention you name so you can verify :+1: 
Anyway, thanks !

PR #2736, opened 2016-08-07
[Feature] Sniping pokemon missing from the pokedex

ChemiHatake:
Was now tested and works fine.


Astifleure:
does someone know if we can configure this bot ?

PR #2740, opened 2016-08-07
Only catch pokemon that are greater CP than current pokemon in bag

theplaymate:
Added a new flag "OnlyCatchHigherCpThanCurrentMinimum". Setting it to true will only catch pokemon that are above CP than your current same pokemon in inventory. Default Value is false.
It helps prevent having so much candy for common pokemon.

### DIFF
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -182,5 +182,6 @@ namespace PoGo.NecroBot.Logic
         SnipeSettings PokemonToSnipe { get; }
 
         bool StartupWelcomeDelay { get; }
+        bool OnlyCatchHigherCpThanCurrentMinimum { get; }
     }
 }
--- a/PoGo.NecroBot.Logic/Inventory.cs
+++ b/PoGo.NecroBot.Logic/Inventory.cs
@@ -256,6 +256,15 @@ namespace PoGo.NecroBot.Logic
                 .FirstOrDefault();
         }
 
+        public async Task<PokemonData> GetLowestPokemonOfTypeByCp(PokemonId pokemon)
+        {
+            var myPokemon = await GetPokemons();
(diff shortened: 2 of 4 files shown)


Taranis01:
In order to prevent high candy ammount for common pokemon wouldn't it need its own PokemonToIngore (List) ? Because the way you explained it right now it would prevent getting enough candys for rare pokemon.

Couldn't we just expand the exisiting PokemonToIgnore list with more values, e.g.:
"pidgey,cp,1000" <-- only better than cp 1000
"pidgey,cp,max" <-- only better than best pidgey cp
"pidgey,iv,96" <-- only better than iv 95
and so on
"pidgey" <-- no pidgeys are catched

theplaymate:
@Taranis01 I wanted to be a bit more generic thats why i made it like this.
Your suggestion is great, I might start working on that as well.

Dano7:
why?  maybe I'm daft, but is there a limit on Candy?  You get Stardust and XP by catching a pokemon even if you transfer it right away.   There is no limit on Stardust, I'm not sure why you would turn on this option, but I guess it is more configurable for those that want to.   Am I missing something?

theplaymate:
@Dano7 for example i get like 500+ Electabuzz candy because they're so common. So people get suspicious by seeing the amount of candy you have. I had a bot who had 2000+ pinsir candy, everyone knew it was a bot. Its for preventing these kind of situations.

daouid:
Only catch pokemon that are greater IV than current pokemon in bag ==> please do the same but for IV

theplaymate:
@daouid will add that feature once this is merged. Anyone please merge so i can do another PR?

Marsinator:
I do believe both are already in the bot if you put keep duplicate to 0

Thaella:
Just use the same filters as the transfer filters but set them in a subroutine for catch filtering.

GrimmGringo:
As Mars stated, this is redundant if you make your settings keep duplicates 0.

PR #2755, opened 2016-08-07
[Enhancement] Missing delay fix, Exp earned on capture...

ChemiHatake:
Now there's a delay when transferring weak Pokemon,
On Pokemon capture it will now output the amount of exp earned, 
and if the capture failed/the Pokemon escaped the output will be colored red.


ChemiHatake:
Can someone please merge/close this? 
I have another idea that I finished and is way more important than this pr.

PR #2775, opened 2016-08-07
Feature: Using gmaps navigation for a more realistic movesimulation

Kotentopf1337:
User needs to provide a GMaps API Key. Get it in the Google DevConsole (its the directions API)


NecronomiconCoding:
What about using OSM ?

Linewalker:
I've made a short test some days ago with gmaps an only per webrequest and without any other plugin. You get everything you need with your private api-key in three lines of code, incl. travel time, meters to endpoint and so on.

FabryB:
Why has this been closed?
We are using Google Elevation anyways for Altitude...

hilycker:
At least write something when you close PRs. This seemed like a good idea.

PR #2858, opened 2016-08-07
[Fix]  Berrys not being used again after Catch Escapes

mrnitrate:
UseBerry was only being used once per pokemon, needs to be reused if the pokemon escapes the catch attempt

### DIFF
--- a/PoGo.NecroBot.Logic/Tasks/CatchPokemonTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/CatchPokemonTask.cs
@@ -45,26 +45,6 @@ namespace PoGo.NecroBot.Logic.Tasks
                     ? encounter.WildPokemon?.PokemonData
                     : encounter?.PokemonData);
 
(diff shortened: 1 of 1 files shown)


GrimmGringo:
I think this should be optional, many people dont want to waste a million berries on a single pokemon. Maybe provide a configurable Max Berries Per Pokemon ?

anadhelf:
sebelumnya saya pake v0.7.1, tapi pas mau relog ada pemberitauan untuk update ke v0.7.2, yaudah saya update tapi malah stuck " log in menggunakan PTC ", setelah saya close. saya coba download langsung v0.7.2, pas mau log in eh ada pemberitauan " data corrupt "...

saran dong

anadhelf:
What this ?

missyyou94:
You should modify that  "UseBerriesOperator": "or",

cawk:
@mrnitrate if you havent got time to add an option to this in the config please let me know as i will do this for you

PR #2896, opened 2016-08-07
Update Inventory.cs - Optimal Number of candies needed for mass evolve

PedroSeda:
Accounts for the candy gained when evolving, can be changed to "-2" if considering transferring the newly evolved pokemon.

### DIFF
--- a/PoGo.NecroBot.Logic/Inventory.cs
+++ b/PoGo.NecroBot.Logic/Inventory.cs
@@ -140,11 +140,11 @@ namespace PoGo.NecroBot.Logic
                     settings.CandyToEvolve > 0 &&
                     settings.EvolutionIds.Count != 0)
                 {
-                    var possibleCountToEvolve = familyCandy.Candy_ / settings.CandyToEvolve;
+                    var possibleCountToEvolve = ((familyCandy.Candy_ - 1) / (settings.CandyToEvolve - 1));
                     amountToKeepInStorage = Math.Max(amountToKeepInStorage, possibleCountToEvolve);
 
                     //remain candy
-                    modFromEvolve = familyCandy.Candy_%settings.CandyToEvolve;
+                    modFromEvolve = (familyCandy.Candy_ - 1)%(settings.CandyToEvolve - 1);
                 }
 
                 var inStorage = myPokemonList.Count(data => data.PokemonId == pokemonGroupToTransfer.Key);


actislav:
what are you doing? open Inventory.cs and find this ` //Lets calc new canBeRemoved pokemons according to transferring some of them and use it as future candy to keepPokemonsThatCanEvolve
                if (modFromEvolve.HasValue)
                {
                    // its an solution in fixed numbers of equations with two variables 
                    // (N = X + Y, X + C >= Y * E) -> (X >= (N * E - C / 1 + E)
                    // where N - current canBeRemoved,  X - new canBeRemoved, Y - possible to keep more, E - CandyToEvolve, C - modFromEvolve(remain candy) ) )
                    canBeRemoved = (settings.CandyToEvolve * canBeRemoved - modFromEvolve.Value) / (1 + settings.CandyToEvolve) + 
                                    Math.Sign((settings.CandyToEvolve * canBeRemoved - modFromEvolve.Value) % (1 + settings.CandyToEvolve));
                }`

and think one more time before fixing something pls

PedroSeda:
I can see the calculation, I just don't understand the need for it. 
I understand the math behind, but it's just too complex for what it needs to achieve.
My simple changes makes that whole block irrelevant, you can see that it never changes the value of **canBeRemoved** because the optimal value was found before.

actislav:
ok, simple example, what if you have 1000 candy in inv. and 12 candy to evolve and 1000 pokemon, with no other requirements to keep besides keepifcanbeevolve, your answer is (1000 - 1)/(12 - 1) = 90.(81)[90], ok you think it normal? by right logic you must keep 1000 / 12 = 83.(3)[83] + possible x pokemon to evolve on candy from transferring 1000 - 83 = 917, 917 / 12 = 76.41(6)[76], but if u just add 83 + 76 u will have 159 pokemon to keep and it equals to 1908 needed candy to evolve, but you will have only 1000 + (1000 - 159) = 1841.... that is why we must to find a solution to two equations with two variables, and you can do it by the formula that i coded for you.

PedroSeda:
But you only need 12 candies to evolve the first pokemon, every other pokemon only needs 11, because every time you evolve you gain 1 candy, to evolve 159 Pokemon you need 159*11+1 = 1750 candies. Your formula gives 153 pokemon to evolve, which leaves 164 candies.

The optimal value would be 166, as my linear programming formulation showed me. I'm working on porting that now. 166 evolves would leave you with 8 candies in the end. Feel free to check my math.

actislav:
"every time you evolve you gain 1 candy" - its new information for me :) ok after work i will check your math, but u right its more simpler calculation, just count that first evolve. Excuse me for my ignorance

PR #2945, opened 2016-08-07
Limit pokemon Lat & Long to 6 decimals

gameuser1:
Limit pokemonCaptureEvent Latitude & Longitude to 6 decimal places in log.


Oolaruu:
Thank you. I recognize one of my issues reported speaking about that. Great

Issue #3024, opened 2016-08-07
Language problem


pacioc193:
i change the language on the config.json but it doesn't affect the bot language. Is a bug?

globus243:
german language seems seems to be broken since 0.7.3 which language are you trying?

pacioc193:
Italian. The bot automatically replace any language into english. i think bug

head0r:
look at the .jsons in the translations dir. every file is available, but open any of them, it is english translation

globus243:
yeah same for me, there is another issue regarding spanish with the same problem.

head0r:
tried to replace the files woth old ones, but NB rewrites them after startup. marking them read-only will cause an error in the app

pacioc193:
No, i have copy an old correct language file, but the bot automatically overwrite it on the start!

Marsinator:
we are aware of that problem and working on a fix right now.

Marsinator:
#2747

ChemiHatake:
was fixed.
the only languages that won't work right now are: vi, cs and ca.
fixes are going to be implemented soon.

Issue #2747, opened 2016-08-07
Language/information issues


Meowrouane:
![image](https://cloud.githubusercontent.com/assets/20751663/17462420/5ed2d5b0-5ca4-11e6-89df-d250613c21a5.png)

*the expression is too long, cannot see the whole informations

*I choosed to keep language EN, yet it started in FR

Oolaruu:
Why not something like : 
Lv: 28 (>>1h08min)
>> of course mean next level in ?

Meowrouane:
the whole 00.00:09:20 thing, 
or maybe force the window to work on full screen instead of half screen, modified log window i mean

Marsinator:
headbar not completely shown on smaller resolutions/screens.
needs improvement

Oolaruu:
Well I think h:min are just right needed.

PR #3044, opened 2016-08-07
don't overwrite translation.LANG.json

somani6:


Marsinator:
#3024 #2747

somani6:
@Marsinator why close? I don't see any other PullRequest that fixes the mentioned problems apart from this one.

PR #3046, opened 2016-08-07
Update and rename PoGo.NecroBot.CLI/Config/Translations/translation.z…

juzhijian:
…h_CN.json to PoGo.NecroBot.CLI/配置/翻译/translation.zh_CN.json


GrimmGringo:
Verify this compiles, it's failing to build.

PR #3096, opened 2016-08-07
added GetTranslation for sniping "unknown" IV pokemon

somani6:
substituted hard coded "unknown" with translatet string
'CommonWordUnknown'


yogyantoro:
Can somebody help me... my sniping cant found pokemon

PR #3161, opened 2016-08-07
Transfer keep logic will now use microsoft expression parser for strimg expression evaluation

actislav:
Backward compatible saved. In future, it wiil help to remove all keep* params from global and per Pokmon TransferFilter and replace it with one string expression with user already defined constants in *old format.

"Accounts for the candy gained when evolving" - commit "6dabf62099c4ca2c6a326797a3eac922e312323f" 
was PedroSeda mistake, candy already accounted, tens line below.
fix reveted


ChemiHatake:
""{0}"" == ""and"" what does that part do?
I tested it and put ""or"" instead of the ""and"" and there was no change

PR #3164, opened 2016-08-08
Prevent IOException from crashing bot, occurs when trying redirecting output stream

weidizhang:
If you try to redirect the output stream of the program and are not actually running it the program directly, you end up getting an IOException saying invalid handle and crashing the bot.

This occurs because of the use of CursorLeft and Clear console functions.

Catching the exception prevents the bot from crashing in the case that the progress bar is not able to load because the output stream is being redirected.


HaiseJr2016:
error

ERROR HELP PLEASE I USE AND THAT WORKS BUT NOW DONT WORK !!! please sorry for mayus  :+1:

PR #3168, opened 2016-08-08
Changed color for CatchEscape and CatchFlee events from red to yellow…

djdookie:
… to be less aggressive. Red should only be used for errors.

catchescapefleeyellow


EthanBianchi:
OH WOW, I just made the SAME exact PR... I'll close mine, but I second this.

djdookie:
Why do you just close this? Red should only used for fatal errors. An escaping pokemon is not an error! Yellow like a warning is much better.

EthanBianchi:
He closed this because as I mentioned before I made the same Pull Request. He must have seen mine first and chose to merge that. It's been implemented.

djdookie:
Yeah i saw that meanwhile. Thx. But you used DarkYellow. I think normal Yellow matched better. ;)

PR #3172, opened 2016-08-08
Fix calculation of EvolveKeptPokemonsAtStorageUsagePercentage to incl…

casselc:
Fix calculation of EvolveKeptPokemonsAtStorageUsagePercentage to include the eggs currently in player's inventory. Without this the player never reaches the correct usage percentage if there are eggs in storage.


ObsessedHacker:
and also fix calculation of IV.

wira93:
Hi I have a question, what's the normal speed in the configuration that I should use, and you guys use? is 31km/hour still realistic? thanks

PR #3176, opened 2016-08-08
Fixed the amount of pokemon to evolve

PedroSeda:
There was too much calculation going on that did not result in the correct value. Spent the whole afternoon checking it. Does not matter whether you transfer a Pokemon or evolve it, you always get 1 candy for it. So the formula is now simplified to reflect it.

### DIFF
--- a/PoGo.NecroBot.Logic/Inventory.cs
+++ b/PoGo.NecroBot.Logic/Inventory.cs
@@ -193,7 +193,7 @@ namespace PoGo.NecroBot.Logic
                     settings.EvolutionIds.Count != 0)
                 {
                     var candiesToUse = familyCandy.Candy_ + inStorage;
-                    var possibleCountToEvolve = (candiesToUse / settings.CandyToEvolve;
+                    var possibleCountToEvolve = candiesToUse / settings.CandyToEvolve;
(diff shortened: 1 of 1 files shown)


ChemiHatake:
was this tested?

PedroSeda:
Yes, been testing for the last 3 hours, no more leftover candies after the mass evolves.

tw88:
I tried it with 0.7.6
https://www.dropbox.com/s/vwrftjp5t54bgvl/Screenshot%202016-08-08%2011.03.34.png?dl=0

config issue? im using 8787 mass xp config

ChemiHatake:
It seems like this has no logic,  why did you add the number of duplicate pokemon to the amount of availavle candies?

dansali:
Same issue as above, even though pokemon can be evolved, they get transfered: http://imgur.com/a/6HqZM

GrimmGringo:

tw88:
it seems that GetPokemonToEvolve creates a list of pkmn to evolve upfront of transfering or evolving. Wouldnt it be more secure to just check if a pkmn is evolvable on transfer? maybe its not so optimized then but maybe it works better

PR #3178, opened 2016-08-08
Don't use red colors for anything but fatal errors!

djdookie:
Use yellow for warnings instead.


djdookie:
Please take Yellow, not DarkYellow, as it optically better matches the green catch messages.

PR #3181, opened 2016-08-08
Code cleanup on Settings, wrapup on logging update

Saiteik:
renamed the logging option that correlates to the changes I made to
better understand its usage. Also moved around and cleaned up the
Settings.


Saiteik:
This is not failing any checks on my end and compiles safely. Please point out my mistake otherwise.

GrimmGringo:
Conflicts and build fails

PR #3186, opened 2016-08-08
Update ConsoleLogger.cs

EthanBianchi:
Simple text "delimiter" to notate a little more clearly what people really want to know.. "What pokemon did I catch?"
The spacing / "*" adds emphasis where it is needed.


GrimmGringo:
Unnecessarily makes the console log even longer

PR #3195, opened 2016-08-08
Get Levelup Rewards

cawk:
Gets the level up rewards automatically


ChemiHatake:
Tested?

cawk:
Was an old piece of code I pushed before but didn't merge due to changes and forgot about it so I just changed the conflicts and re pushed and it worked when I made it can test now as I'm level 36 and takes ages to level up

djdookie:
Take a new lvl 1 account to test then. ;)

cawk:
Why don't you?
This code worked before there are no significant changes to what I had to do so I am 98% sure it will still work

PR #3201, opened 2016-08-08
Randomized Recycle

KOTR2003:
More human-like recycling. Adds a random variable to the 'amount to keep' for each of the items. Random value is between -5 and +5.

Example: Max potions to keep is set to 50, it will keep between 45 and 55.

... still failed the check, I give up (clearly I am not a programmer haha)


GrimmGringo:
This does not do what the user wants; they're setting it to a specific value and expect it to be that value. For that reason we cannot pull this in, thank you though!

KOTR2003:
In the config, you can have a boolean value to take this into account or not. Trashing everything to the exact same number of item count is not humanlike. Expecially if you have 3/4 different pokeballs. Who sits there and adds up those multiple item counts to trash to an exact number total?

Just figure this is an easy catch to tell if a bot is recycling items.

PR #3205, opened 2016-08-08
Add alert for Pokémon type, CP or IV

PiscisSwimeatus:


GrimmGringo:
I personally don't see the need for this?

cawk:
neither do i 
why would someone want to use this?

rahulrverma:
Alert sound for errors or when the bot asks to key press would be a better addition. :)

PiscisSwimeatus:
Well I think its a good addition hence why I made it ;) It makes it so much easier to catch those IV 100 without needing to watch the console all the time.

NecronomiconCoding:
After a discussion with the Dev Team we have to decline that PR

PR #3212, opened 2016-08-08
PokemonCatch event logging cleanup

Saiteik:


Energex22:
There should be options of what information to show, like me personally i don't really care the location of the pokemon, especially if i am not sniping at them, even then not so much.

PR #3218, opened 2016-08-08
Pokestop event logging cleanup

Saiteik:


ChemiHatake:
You'll need to change it in all the languages.

EthanBianchi:
Are you planning to do it in all languages, or should someone else consider that a "to-do"?

Saiteik:
I will make the change for all translations however someone native to that language will need to update it specifically.

PR #3243, opened 2016-08-08
Use berry when pokemon escape and choose how many berries to use per pokemon

cawk:
As title says have tested and works fine


cawk:
used the berry for each time if they escape from @mrnitrate
i just added the how many berrys to use on a pokemon

cawk:
Just renamed config names for better understanding

cawk:
fixed a debug error i made

ThomasSRL:
Wont this break the softban recognition?
Anyway thanks for adding new features, I just dislike the fact that it always seems to expand the config.

cawk:
no it wont break it because catch escape and catch flee are different things

PR #3253, opened 2016-08-08
PTC Login fixed, Pokezz.com fixed

d-moos:
PTC login was stuck (crashed -> proxy misconfiguration).

http://www.pokezz.com Sniper is working again.
using Socket.IO client now (quick n dirty fix, sorry).

sniping still got some debug messages and not working 100%. ill check it when im back home.


Omigof:
pokezz still not work for me

Issue #3223, opened 2016-08-08
[Bug] Info Spam


djdookie:
I see this all the time. Seems to be a bug in the state machine?

![bugmultiinfo](https://cloud.githubusercontent.com/assets/17240204/17483845/ecac196c-5d87-11e6-8448-7d3d117cba14.PNG)

I use the current version v0.7.6, but I see this since multiple releases.

ObsessedHacker:
@djdookie Same Issue.

EthanBianchi:
I think it's just constantly checking if you have enough pokeballs for sniping, and updating you every second while you visit pokestops so it knows when to stop.

I'll take a peak at the code, maybe we can settle down the info loop.

Energex22:
Seen it a couple times myself, but not recently.
Wonder what set's it off doing that.

mo0ojava:
I'm having the same or perhaps a similiar problem. There's constant <INFO>-spam when using a gpx-track. Otherwise no spam. The wiki states that "If UseGPXPathing is set to true and GPXFile is found, Evolution, updating, transferring, recycling and sniping will execute every second or based on the value in milliseconds for MinDelayBetweenSnipes (whichever is higher)."

Attaching the config and a pic of the phenomenon.

![clp](https://cloud.githubusercontent.com/assets/20909700/17487733/70045168-5d96-11e6-8fa1-d45f0fce126b.png)
[config.json.txt](https://github.com/NECROBOTIO/NecroBot/files/407259/config.json.txt)

nfettinger:
How long is your GPX file. I assume it prints this out every time it starts over in the file.

mo0ojava:
I created a rather "clean" gpx on WTracks just to try to eliminate the chance that it was gpx-file-related. Attaching the testfile. The file seemed to parse well when surveyed through the visualizer. Afaik the bot works great. It just happens to spam.. a little.

I'd estimate that the text above starts repeating every 2-3 sec after it has finished dumping the previous one.

[gruvan.gpx.txt](https://github.com/NECROBOTIO/NecroBot/files/407312/gruvan.gpx.txt)

EthanBianchi:
I think I found an easy fix for this, does anybody that uses GPX want to test my fix? (I don't use it)

mo0ojava:
Sure, I'd gladly try!

EthanBianchi:
It seems too easy to fix it, but worth a shot imo

djdookie:
[...]

PR #3277, opened 2016-08-08
[Needs Testing] Fix for GPXPathing Log Spam

EthanBianchi:
#3223 Lots of log spam occurs while using GPX Pathing, and from what I can tell this should significantly cut down on the spam in the log.

### DIFF
--- a/PoGo.NecroBot.Logic/Tasks/FarmPokestopsGPXTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/FarmPokestopsGPXTask.cs
@@ -99,15 +99,10 @@ namespace PoGo.NecroBot.Logic.Tasks
                                     Latitude = pokeStop.Latitude,
                                     Longitude = pokeStop.Longitude
                                 });
-                            }
-                            else
-                            {
-                                await RecycleItemsTask.Execute(session, cancellationToken);
-                            }
-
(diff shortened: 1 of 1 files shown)


mo0ojava:
This regards the problem described [here](https://github.com/NECROBOTIO/NecroBot/issues/3223#issuecomment-238313672).

GrimmGringo:
Check your code, looks like you're missing a }

djdookie:
Does it also help with the info spam without GPX-Pathing? I am not using an GPX file, but also have this info spam, like mentioned in #3223. Same reason?
Thanks for your help.

EthanBianchi:
I can't figure out where I'm missing a bracket lol. I can build/compile just fine, and I've tried following the scope as best as I know how... Can't seem to find what I'm missing lol.

PoGo.NecroBot.Logic/Tasks/FarmPokestopsGPXTask.cs
```diff
@@ -213,4 +208,4 @@ private static List<GpxReader.Trk> GetGpxTracks(ISession session)
             return pokeStops.ToList();
         }
     }
-}
```
djdookie: Possibly this bracket is missing @EthanBianchi?

PR #3292, opened 2016-08-08
Randomize Recycle v2

KOTR2003:
Changed the random recycle from the first version to allow for users to configure it on or off, and how random it will be.

Purpose for this is to make recycling more human-like.
Example:
Pokeballs: 73
Greabballs: 49
Ultaballs: 31
Masterballs: 11
Total pokeballs = 164
TotalAmountOfPokeballsToKeep set to 120

For a human to recycle to 120, they would need to add the quantity of each of their pokeballs up and scrap accordingly. This is highly unlikely to happen, and more often than not they would just scrap a random amount of the weakest pokeballs. 

The config will give an option to turn if on and off, and set the +/- randomized value to use.


KOTR2003:
Haha, I noticed that ILogicSettings afterwards >.< you got to it before me

PR #3296, opened 2016-08-08
[BUG FIX] Fixed evolution mistakes/bugs.

ChemiHatake:
Now evolving using lucky eggs and keeping Pokemon that can evolve
both work.


ChemiHatake:
[Do not merge, I need to test one more thing and unfortunately ptc is down]

dansali:
I also noticed that when it says I have 10 evolvable pokemon, it will only evolve 5. Seems to always be half, is this fixed in this revision?

nfettinger:
I believe this is fixed. Open a new issue with actual log details if not.

PR #3304, opened 2016-08-08
Pokemon trasfer readability

FabryB:
Improve text formatting with padding for known fields length in order to
improve readability of transfer log


FabryB:
BEFORE:
before

AFTER:
after

PR #3320, opened 2016-08-08
Fix PokemonUpgrade by IV task

Saiteik:


Saiteik:
The changes made on the Pokestop event will need to be done as a "To Do" for other languages.

ChemiHatake:
You will need to change it in all the languages
to minimize errors

Saiteik:
Leave a comment

PR #3329, opened 2016-08-08
Idea/Not for release: Professor_Kukui rename Scheme

PiscisSwimeatus:
My dirty rename script. Just posted it as is. It might be something you find interesting and want to rewrite/include in your bot. Its based on [(https://www.reddit.com/r/TheSilphRoad/comments/4wkqde/perfect_iv_pokemon_go_species_data_w_movesets/](url) Professor_Kukui move set calculation. It uses the following format: AAbCC-DDEEFF where:

AA = Inverted percentage of the best obtainable pokemon/move set for attack. ie. 1 = best, 17 = 17 % less effective than #1, 49 = 49 % less and etc.
b = Best attack: p=Primary s=Secondary
CC = Same as AA but for defending
DD = Individual Attack (IV stat)
EE = Individual Defense (IV stat)
FF = Individual Stamina (IV stat)


ChemiHatake:
actually it is pretty useful, 
but maybe it shouldn't replace the current RenamePokemonTask.

PiscisSwimeatus:
Yeah its my personal preference hence its all hard coded. For public release you prob want to split each element into a part of RenameTemplate for the user to choose.

GrimmGringo:
This cannot be merged in until it's optional to keep the current method, OR use your method.

PR #3445, opened 2016-08-09
pt-Br Update

CoderHaxSrubz:
Just As Simple, i noticed that pt-PT and pt-BR traslitions are the same, so i made an pt-BR translation for the Brazilian Portuguese Speakers

### DIFF
--- a/PoGo.NecroBot.CLI/Config/Translations/translation.pt-br.json
+++ b/PoGo.NecroBot.CLI/Config/Translations/translation.pt-br.json
@@ -2,15 +2,15 @@
   "TranslationStrings": [
     {
       "Key": "pokeball",
-      "Value": "PokeBall"
+      "Value": "Poké Bola Normal"
     },
     {
       "Key": "greatPokeball",
-      "Value": "GreatBall"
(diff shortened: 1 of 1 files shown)


GUiHKX:
I think your translation file is very inconsistent.
* There are some words missing proper accentuation; `Pokébola` is better than `Poké Bola`, because in english it's only one word too;
* Uncessary capitalized words in the middle of a message;
* On line 33, you completely changed the meaning of the error message, there's nothing to do with offline servers;
* Overcomplicate the meaning of the message (you forgot to translate the rest of it, by the way): `O Email e/ou A Senha esta(ao)`, when you can simply translate to: `E-mail e senha não coincidem.`;

But don't get me wrong, I don't want to discourage you from keep improving the translation, I just think you can do better. I liked that you reverted some words back to english, because we still don't have the "official" translation for those.

Thanks for your effort.

CoderHaxSrubz:
My translation its more based on colletive user experience, so in line 33 where i meant to say its that, in the official discord group when the servers get down , evryone gets the line 33 message. Learned The tips and will figure it out!

PR #3501, opened 2016-08-09
Add back a working Pokesnipers sniping implementation

jjskuld:
This bypasses the cloudflare protection on Pokesnipers api.  Also introduce new config option "GetSniperInfoFromPokeSnipers": "true".  This is OR'd with the Pokezz snipers, so you can snipe from both sites at the same time, or disable one or the other as wished.


jjskuld:
Just need to set in config:
```
"GetSniperInfoFromPokeSnipers": true,
```

Also to grab more pokemon, set:
```
"SnipeIgnoreUnknownIv": false,
```
because Pokesnipers does not always return a valid IV.

Here is a screenshot of sniping from Pokesnipers.

sniping from pokesnipers

Issue #3455, opened 2016-08-09
v0.7.7 logging in issue


taihangy:
I have an issue I can't logging in using v0.7.7 since this afternoon.

djdookie:
Same here, Login working in v0.7.6 hangs at "(INFO) Logging in using Ptc" in v0.7.7.
Guys you should really test your "optimizations"! ;)

vfuyong:
Set LoginSuccessProbability = 1

djdookie:
Ok this seemed to be a problem with the servers.
After debugging the code, (for me) the reason was that no ptc authtoken was returned.
Unfortunately the client just said nothing, didn't catch the correct exception and just tried again 20 seconds later via default exception case. No message, no logging. Very bad for the user! ;)
I will try to optimize this behaviour. But this failure is hard to test because meanwhile the servers work fine again. ^^

PR #3538, opened 2016-08-09
Ptc login problem optimization

djdookie:
Trying to solve #3455 
If the ptc login didn't return an ptc authtoken, the user now get's the according feedback and the client should retry after a short time.
Before there was no message, no logging and a standard retry interval of 20 seconds, because no correct exception were caught.
This looked like the application just hanged for the user.

### DIFF
--- a/PoGo.NecroBot.Logic/Common/ApiFailureStrategy.cs
+++ b/PoGo.NecroBot.Logic/Common/ApiFailureStrategy.cs
@@ -48,7 +48,7 @@ namespace PoGo.NecroBot.Logic.Common
         {
             try
             {
-                if (_session.Settings.AuthType != AuthType.Google || _session.Settings.AuthType != AuthType.Ptc)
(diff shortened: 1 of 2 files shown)


djdookie:
I did a small modification in submodule FeroxRev/RocketAPI. Build seems to fail because of this.
How can I fix this?

Meeplez:
heres a better one
my pullz r fugged



`            try
            {
            begin:
                await Task.Delay(1000);
                if (_session.Settings.AuthType != AuthType.Google || _session.Settings.AuthType != AuthType.Ptc)
                {
                    await _session.Client.Login.DoLogin();
                }
                else
                {
                    goto begin;

                  //  _session.EventDispatcher.Send(new ErrorEvent
                 //   {
                    //    Message = _session.Translation.GetTranslation(TranslationString.WrongAuthType)
                 //   });
                }
            }`

GrimmGringo:
Implemented manually, will commit momentarily

djdookie:
Thanks. But you should please implement the changes in submodule FeroxRev also, as it is part of my change.

PR #3554, opened 2016-08-09
Merge

evoamac:
Recycle when inventory is full
Fix null reference error with null check

### DIFF
--- a/PoGo.NecroBot.Logic/Inventory.cs
+++ b/PoGo.NecroBot.Logic/Inventory.cs
@@ -360,18 +360,28 @@ namespace PoGo.NecroBot.Logic
         {
             var inventory = await GetCachedInventory();
 
+
             var families = from item in inventory.InventoryDelta.InventoryItems
-                            where item.InventoryItemData?.Candy != null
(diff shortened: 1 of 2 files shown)


BryanSuero:
` if (++stopsHit >= storeRI || fortSearch.Result == FortSearchResponse.Types.Result.InventoryFull) //TODO: OR item/pokemon bag is full //check stopsHit against storeRI random without dividing.
                  {		                  {
                      storeRI = rc.Next(6, 12); //set new storeRI for new random value		                      storeRI = rc.Next(6, 12); //set new storeRI for new random value
                      stopsHit = 0;		                      stopsHit = 0;
 -                    if (fortSearch.ItemsAwarded.Count > 0)		 +                    if (fortSearch.ItemsAwarded.Count > 0 || fortSearch.Result ==FortSearchResponse.Types.Result.InventoryFull)`

I just think you don't have to check twice :+1:

evoamac:
because some time the inventory is full
and when inventory item is full fortSearch.itemsAwarded.count = 0. thus the recycle won't execute.

PR #3559, opened 2016-08-09
Fixing some of the translation.id.json

kadektop2:
Readjusting some words to make it more sensible.

PR #3568, opened 2016-08-09
Bahasa Indonesia: update translation.

awaaas:
`Kredensial` turns out to be not a word in Indonesian. So,  will be using `detail pengguna` from now on out. It's easier to understand too.


awaaas:
Also turns out to be fixed by #3559. Closing this.

PR #3606, opened 2016-08-09
Speed tweaks

GrimmGringo:


PoGo.NecroBot.Logic/State/FarmState.cs
```diff
@@ -19,43 +19,43 @@ public class FarmState : IState
...
+            if (session.LogicSettings.TransferDuplicatePokemon)
             {
-                await LevelUpPokemonTask.Execute(session, cancellationToken);
+                TransferDuplicatePokemonTask.Execute(session, cancellationToken);
             }
 
             if (session.LogicSettings.UseLuckyEggConstantly)
             {
-                await UseLuckyEggConstantlyTask.Execute(session, cancellationToken);
+                UseLuckyEggConstantlyTask.Execute(session, cancellationToken);
```
khellang: Since these tasks are no longer awaited, you'd get unobserved task exceptions if something goes wrong. That seems unfortunate. Was this intended?

N0ct3:
Not sure the "speed" increase is worth the risk of something going horribly bad without us noticing it...

khellang:
Well, in pre-.NET 4.5, a `UnobservedTaskException` would bring down the process, but with .NET 4.5+, it'll just get swallowed silently. You can, however, observe the task through the [`TaskScheduler.UnobservedTaskException`](https://msdn.microsoft.com/en-us/library/system.threading.tasks.taskscheduler.unobservedtaskexception(v=vs.110).aspx) event. I suggest you either handle this event and log something, or simply keep a reference to all the tasks (in a list or something) and return a `Task.WhenAll` task from the method.

PR #3729, opened 2016-08-09
 fix v0.8.0 speed tooooo fast cause got a lot of softbans

Greatliu89:
Set "DisableHumanWalking" to False

speed up can change "WalkingSpeedInKilometerPerHour" allright?
the v0.8.0 default is teleport not walking


Greatliu89:
"WalkingSpeedInKilometerPerHour" can under 51

noodle-:
Horrible title & description, but valid change/commit. :+1:

thecantero:
So this is the culprit. thanks

PR #3806, opened 2016-08-09
Added IV value to /top and /all in TelegramService

Plotin:
I think it can be convenient to see the IV perfection value in telegram also


GrimmGringo:

Plotin:
Another commit already added IV values, closing this one

PR #3836, opened 2016-08-09
[FIX] No Delay on Recycle

KOTR2003:
The delay between actions was removed for recycling items, causing all the actions to be performed within milliseconds of each other, essentially insta-recycling all items at once.

If they detect this, it is a huge red flag saying you are a bot. 

Here I have only added it back in, but if needed the config could have a boolean option for this to allow the user to disable the timer

### DIFF
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -164,6 +164,7 @@ namespace PoGo.NecroBot.Logic
         bool SnipePokemonNotInPokedex { get; }
         bool RandomizeRecycle { get; }
         int RandomRecycleValue { get; }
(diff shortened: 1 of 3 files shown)


GrimmGringo:
Add a "DelayAfterRecycling" option, set to false by default, and use that to determine whether to delay them. Recycling items can really slow the bot down.

LilBifi:
I strongly believe the missing delay between recycling items is the reason so many are complaining about softbans with v0.8.0.

helooguys:
take a look at that https://github.com/NECROBOTIO/NecroBot/issues/3857 . Its more like cosmetic bug if i can say so.

LilBifi:
Try setting `"VerboseRecycling": true` and you'll notice the bot tries to recycle all items at once.

PR #3887, opened 2016-08-09
Upgrade to .NET 4.6.2

JamieMagee:
.NET 4.6.2 is supported on Windows 7 and above. This would only exclude Windows Vista, as XP doesn't support 4.5.2.

### DIFF
--- a/PoGo.NecroBot.CLI/App.config
+++ b/PoGo.NecroBot.CLI/App.config
@@ -1,31 +1,31 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <configSections>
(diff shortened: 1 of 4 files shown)


JamieMagee:
Appveyor build fails due to it not having .NET 4.6.2 installed. They targeted it for 7th Aug (See [here](https://github.com/appveyor/ci/issues/960)) so it should be available soon.

GrimmGringo:
We aren't going to approve something that's going to cause every build going forward to fail in appveyor.

JamieMagee:
I agree, but I feel that there is no harm in leaving the PR open for a day or two until AppVeyor update their environment.

jcapellman:
Awaiting for this update as well - now using 4.6.2 across the board for my latest project.

PR #3895, opened 2016-08-09
Basic plugin architecture.

tstiegler:
## What is this?

This addition allows for basic plugins to be created for NecroBot. These plugins are DLLs can be dropped into the Plugins directory that is now created in the NecroBot running folder. At runtime, these DLLs are loaded and every class that implements `INecroPlugin` are instantiated. The console, necrobot session, settings and statistics are exposed to any plugin that is instantiated 

### Why?

I have been working on implementing my own html5/websockets fontend for NecroBot and wanted to implement my own websockets server. This meant polluting the CLI project with a bunch of code that probably didn't need to be there. This way, people can create extensions to necrobot and share them without polluting the codebase.

### What are the modifications to ConsoleLogger for?

I wanted to be able to make anything that is logged out to console by NecroBot able to be lisened in on by my frontend UI. This meant making the `ConsoleLogger` available to any plugins, as well as creating a event that can be hooked into, to be able to receive and forward anything that is logged to console.

### How do I create a plugin?

Just create a new class library project outside of NecroBot, add in these references:

- NecroBot.exe
- POGOProtos.dll
- Pokemon.Go.Rocket.API.dll
- PoGo.NecroBot.Logic.dll

And then create a class that implements `INecroPlugin` that does what you want it to. Compile this code to a DLL and then drop this into the Plugins folder of NecroBot. Running NecroBot will automatically instantiate the class and run the Initialize method.

You can see an example of this code here: https://github.com/tstiegler/nurx-websockets-backend
And see a video of it in action here:

- https://puu.sh/qvcA6/1dbb27c6f1.mp4 (Horrible video FPS, sorry)
- https://i.imgur.com/JnDpXeZ.gif (Better video, but only shows frontend which is unrelated to this PR)


ChemiHatake:
I vouch for this

GabGonzalez:
i sup on these one

ianpugh:
Great idea!

TBulbaDB:
I would say, instead of dll's, adding C# files and bot compiling them at launch, you never know what u'll get inside a DLL.

ianpugh:
@TBulbaDB , then it's not a true plugin architecture. You could still do that with this via the interface.

TBulbaDB:
No, check Demonbuddy for example, u use the IPlugin interface in your code and put all your code in a subfolder in Plugins folder, bot detects that code during startup and compiles it using Roslyn.

ianpugh:
@TBulbaDB Okay -- that makes sense. I bet once this is merged you could make a PR to include those changes as an addition. That'd be a good feature.

TBulbaDB:
Ah, I have no time to work on that, just a suggestion :)

tstiegler:
@TBulbaDB Either way, the average user would be blindly trusting a lot of stuff that they don't understand. It goes without saying, if you don't trust it, don't install it. If you don't trust it and it's open source, you can comb through the source and compile it yourself to be safe... Just like NecroBot itself.

PR #3945, opened 2016-08-09
Fix pokezz sniping socket bug

jjskuld:
Occasionally, you can get an error adding new pokemon to collection when it is enumerating the list.  The socket should be closed immediately after pokemon data is received to minimize the chance of this occurring.


jjskuld:
Let me rebase and resend

Issue #3833, opened 2016-08-09
Telegram pokédex request


spylor:
"/pokedex" would be a useful command in telegram to see what pokémon you are still missing/already have.

PR #3956, opened 2016-08-09
Telegram: Added new /pokedex Command

MineMalox:
Changelog:
* Added /pokedex Command (See #3833)


GrimmGringo:
New language keys have to be added to every single language, with English text so they can be translated. Please resubmit once this is done.

PR #4015, opened 2016-08-09
Only use good eggs in limited incubators

ldevil-ch:
Added minEggKm for limited incubators to avoid using limited incubators on anything but 10km eggs (settings can be changed to use also 5km eggs or all eggs)

### DIFF
--- a/PoGo.NecroBot.Logic/ILogicSettings.cs
+++ b/PoGo.NecroBot.Logic/ILogicSettings.cs
@@ -89,6 +89,7 @@ namespace PoGo.NecroBot.Logic
         bool TransferDuplicatePokemon { get; }
         bool TransferDuplicatePokemonOnCapture { get; }
         bool UseEggIncubators { get; }
+        int minEggKmForLimitedIncubators { get; }
         int UseGreatBallAboveCp { get; }
         int UseUltraBallAboveCp { get; }
         int UseMasterBallAboveCp { get; }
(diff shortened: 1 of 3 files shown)


benda95280:
Could we prioritize it ? Because if no more 10kms ... :(

DBa2016:
Don't get me wrong, but why exactly would one want to hatch only 10km eggs in limited incubators? Do they guarantee stronger pokemons? Because if not, it is of no difference whether you hatch a 2km or a 10km egg. Hatching a 2km one frees up space quickly and gets you some XP, a 10km one gets you a better pokemon but takes much longer...

benda95280:
So you can read to : 
"Could we prioritize it ? Because if no more 2kms ... :("

ldevil-ch:
Well this only helps if you do not have an infinite amount of limited incubators :) When you put only 10km eggs in the limited ones you will on avarage walk further with two (or more) active incubators.
As the bot will really quickly "walk" all 2km eggs in the unlimited incubator they should not be put into the limited ones. While you walk the 3 use 10km you can walk 15 2km eggs in the unlimited one which would need 5 limited incubators.
With 5km eggs it is less difference that is why you can also set it to 5 to allow also 5km eggs to be put into the limited incubators. If you do not wish to use the setting, you can simply put it to 2 and all eggs will be used.

GrimmGringo:
Change the default to 2.

PR #4022, opened 2016-08-09
Ordered pokemon transfer list Lowest->Highest

Andrerm124:
This should make it easier for people to react to unexpected pokemon being transferred (Due to config issues etc).


benda95280:
you're the king ;)

PR #4033, opened 2016-08-09
Change Level Up Colour

cawk:
Changed levelup colour to something different seemed to bland to be the same as the others


PoGo.NecroBot.Logic/Logging/Logger.cs
```diff
@@ -114,5 +114,6 @@ public enum LogLevel
         New = 15,
         SoftBan = 16,
         Debug = 17,
+        LevelUp = 18,
```
TheBorna: LevelUp = 17
Debug = 18

As right now debug messages will show up :P

cawk:
thank you i forgot to push my commit with that fix 
just tested it works fine

PR #4077, opened 2016-08-09
Add teleport option - 

jjskuld:
Instead of using human-like walking, this option allows you to teleport directly to pokestops.  You will get more softbans using this, but it can greatly speed up the xp/hr even with softbans.

Disclaimer: By the way, I did look at code from the competing fork of this bot.  So credits should go to the other bot.  The teleport code in the other bot was a little messy, so the bare minimum code was implemented to keep the behavior of teleporting and walking as close as possible.


dddbliss:
Can you disclose the original source to credit please?

jjskuld:
Original source is from PokeMobBot.

NecronomiconCoding:
We already have Teleport... :lollipop:

jjskuld:
Ok, I see it now.. ```DisableHumanWalking``` is teleport.  I'll close this out then.

PR #4118, opened 2016-08-09
[now automatic] instantly update config.json file changes

msx752:
we dont have to restart, when we changed some options, now automatic
#IMPORTANT (you make sure access to config.json, otherwise you may get errors)

### DIFF
--- a/PoGo.NecroBot.Logic/State/StateMachine.cs
+++ b/PoGo.NecroBot.Logic/State/StateMachine.cs
@@ -5,6 +5,8 @@ using System.Threading;
 using System.Threading.Tasks;
 using PoGo.NecroBot.Logic.Event;
 using PokemonGo.RocketAPI.Exceptions;
+using System.IO;
+using PoGo.NecroBot.Logic.Logging;
 
 #endregion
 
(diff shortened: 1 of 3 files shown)


PoGo.NecroBot.Logic/State/StateMachine.cs
```diff
@@ -29,6 +31,24 @@ public void SetFailureState(IState state)
...
+            configWatcher.Filter = "config.json";
+            configWatcher.NotifyFilter = NotifyFilters.LastWrite;
+            configWatcher.EnableRaisingEvents = true;
+            configWatcher.Changed += (sender, e) =>
+            {
+                if (e.ChangeType == WatcherChangeTypes.Changed)
+                {
+                    session.LogicSettings = new LogicSettings(GlobalSettings.Load(""));
+                    configWatcher.EnableRaisingEvents = !configWatcher.EnableRaisingEvents;
+                    configWatcher.EnableRaisingEvents = !configWatcher.EnableRaisingEvents;
```
khellang: Why is this line appearing twice?

msx752:
@khellang because event's data is coming twice(echo) so i need reset configWatcher

first EnableRaisingEvents  is false (stop watching)
second EnableRaisingEvents  is true (start watching)

so 2nd data doesn't come

PoGo.NecroBot.Logic/State/StateMachine.cs
msx752: @khellang because event's data is coming twice(echo) so i need reset configWatcher

first EnableRaisingEvents  is false (stop watching)
second EnableRaisingEvents  is true (start watching)

so 2nd data doesn't come. i didnt found alternative or i dont now why data comes twice

PR #4143, opened 2016-08-09
Adding proxy check

GrimmGringo:


LordDeimos:
@GrimmGringo speaking about the issue #4204 , is there a problem raising the timeout into NecroWebClient from 5000 to 10000 ?

Issue #3513, opened 2016-08-09
have the bot auto restart after auto updating


Marsinator:
right now it just closes down and we will have to manually start it again after the update.

schwartzadev:
Wouldn't this require a daemon to restart the bot?

KralElite:
make sure there is a way to update and close.  Solution :Make a timer countdown till restart so user can close if unwanted restart

PR #4157, opened 2016-08-09
Bot now restarts when updating

Andrerm124:
If the bot has updated, successfully transferred a config, and the user has not chosen to open the config; the bot will restart automatically.

Solves issue #3513


Simpuhl:
Wait, so it will ask us to double check config first before auto opening right?

Sometimes things get changed accidentally so i like to check it before I proceed

djdookie:
You could make that configurable.

PR #4207, opened 2016-08-10
Bug fix - Clean up sniping logic

jjskuld:
1. Remove redundant checks to locations visited (was checking this twice unnecessarily)
2. Remove double-addition and incorrect addition to locations visited (it would blacklist a location even if sniping did not occur due to full pokemon inventory for instance).
3. Cleanup last sniping time (should be set within the ```Snipe``` function).

By the way, this is a revert of most of the changes made in 329306bf469437827376204f923234f059f66be7


djdookie:
Nice, thanks for cleaning up the snipe code!

PR #4124, opened 2016-08-09
Telegram: Added new /pokedex Command

MineMalox:
Changelog:

Added /pokedex ( See #3833)
Fixed /top \<cp/iv\>\<amount\>

PR #4211, opened 2016-08-10
Bahasa Indonesia: fix and update translations.

awaaas:
New untranslated string (from #4124).
`Koordinat` is a place, so must be separated from `di`, @wakrad.


kadektop2:
welp I missed them

PR #4251, opened 2016-08-10
Add Ability for Missed Throws

KOTR2003:


KOTR2003:
Says it has errors, but it checked out of order. The code has been tested and works fine (several hour run).

GrimmGringo:
You're going to need to resubmit it after verifying that the build passes, because we cannot accept a merge request with appveyor failing.

PR #4441, opened 2016-08-10
Update PoGo.NecroBot.CLI.csproj

NunoAG:
https://github.com/NECROBOTIO/NecroBot/issues/4421


cawk:
The Encrypt.dll is needed for the api

PR #4480, opened 2016-08-10
Add Ability for Missed Throws

KOTR2003:


cawk:
Do you want me to do this for you seems you are having some issues?

KOTR2003:
@cawk Revert the changes, just realized there was another file to modify too... just hadn't committed it, and of course I am at work and don't have Visual installed on here...

cawk:
yeah you have to commit the api also 
just PR when you fix it ive already tested your code so if any dev notices it just merge it

KOTR2003:
Ah ok, I see you added the 'HitPokemon' to the API, thanks! 

It should all be good then!

cawk:
no i reverted that commit 
since i didnt realise your PR was closed just make a PR to both when you are ready

KOTR2003:
The API still shows the edit committed which is fine. Will (attempt) to make a PR, in before appveyor build errors.

PR #4486, opened 2016-08-10
German translation corrected, inventory cleanup corrected.

DBa2016:
## Short Description:
- updated Telegram help summary in German translation
- when inventory full, included transferring weak pokemons (...if enabled in config file)


DBa2016:
I would love to do this, but don't know how... Do I need to create an extra branch on my fork?

PR #4491, opened 2016-08-10
Upgrade to .NET Framework 4.6.2

JamieMagee:
## Short Description:

Upgrade from .NET 4.5.2 to .NET 4.6.2. .NET 4.6.2 is supported on Windows 7 and above. This would only exclude Windows Vista, as XP doesn't even support 4.5.2.

### DIFF
--- a/PoGo.NecroBot.CLI/App.config
+++ b/PoGo.NecroBot.CLI/App.config
@@ -1,31 +1,31 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <configSections>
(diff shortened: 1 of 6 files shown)


akkatracker:
What benefits does the upgrade provide? Simply curious - I don't deal a lot with .NET

Sakuto:
Actually nothing, but in the future commit it would allow people to take profit of the new features of the .NET framework.

JamieMagee:
@akkatracker @Sakuto .NET 4.6.2 brings in some nice debugging features, but the biggest changes are in .NET 4.6. See [here](https://msdn.microsoft.com/en-us/library/ms171868(v=vs.110).aspx#v46) for a complete list of changes.

Issue #4239, opened 2016-08-10
let us use more than 1 moveset to keep pokemon


Marsinator:
so as of right now we do have movesets added to the pokemontransferfilter (THANK YOU).

however, after talking to @dddblix I was made aware, that it is currently "only" possible to keep pokemon that have either one of two moves (operator or) or both of these moves (operator and).

now theres several moves a pokemon can have, you might already know this. this results in multiple possible movesets. Seeing more and more people reach level 40 they will want to optimize their gym lineup for best movesets.

the point im trying to make is that theres more than one "best moveset", because sometimes you attack, sometimes you defend a gym.

this chart might help explaining why I will ask the following: https://embed.gyazo.com/28dfb36c34439bf8862a149e41369203.jpg

can we have more than one moveset for us to keep with operator "and"?

like this:
: [["onefast","two"],["threefast","four"]]?

so that it keeps pokes which have either onefast&two (best attack moveset) OR threefast&four (best defensive moveset).

Just to be clear I'll add Dragonite example of what I would like to see:

 "Dragonite": {
      "KeepMinCp": 2600,
      "KeepMinLvl": 0,
      "UseKeepMinLvl": false,
      "KeepMinIvPercentage": 95.0,
      "KeepMinDuplicatePokemon": 5,
      "Moves": [
        ["dragonBreathFast",
        "dragonClaw"],
        ["steelWingFast",
        "dragonPulse"]
      ],
  "KeepMinOperator": "or",
      "MovesOperator": "and"
    },

xafelld:
gogo high-tech transfer filters!!

junfox02:
can you put whole config. cuz i dont understand. where do i put this?

Rabelaiss:
I will also love this feature, because IV is less important than moveset, i prefer a 85 iv best moveset than a 100 iv shiet moveset

PR #4581, opened 2016-08-10
Allow multiple movesets to be defined.

dddbliss:
## Short Description:

Allow multiple movesets to be defined per pokemon.

```
"Moves": [
   ["QuickAttack","AirCutter"],
   ["Bash","AirCutter"]
]
```

*Note*: Might cause on update the loss of old Move settings. This is due to new data structures that are not transferable.

## Fixes (provide links to github issues if you can):
#4239


dddbliss:
MoveOperator still works, and and/or or still works between the moves in a individual set as before.

MoveOperator "or":
"QuickAttack" or "AirCutter"

MoveOperator "and":
"QuickAttack" and "AirCutter"

PR #4690, opened 2016-08-10
Added all pokemon to pokemon list

Markersx:
why are not all the pokemon on the list?


Markersx:
@GrimmGringo can i get an explanation?

PR #4705, opened 2016-08-10
Fixed transferring ALL pokemons if no moves in config

somani6:
the ForEach didn't got triggered at all if no Moves are defined.
I fixed that and set the default return value to false in case something
else goes horribly wrong at least the pokemon shouldn't be transfered.

### DIFF
--- a/PoGo.NecroBot.Logic/Inventory.cs
+++ b/PoGo.NecroBot.Logic/Inventory.cs
@@ -103,18 +103,29 @@ namespace PoGo.NecroBot.Logic
                         {
                             var pokemonTransferFilter = GetPokemonTransferFilter(p.PokemonId);
 
-                            bool ret = true;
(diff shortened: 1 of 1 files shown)


actislav:
there is in your commit logic mistake, i mean in author of multilist, i will fix it, ret not mereged and will used last one if movelist have 2 or more list in it

somani6:
@actislav feel free to correct me :) i'm not that into c# but it seems to work for me at least :)

jjskuld:
This needs to be merged ASAP.  Current master will transfer all your pokemon, ignoring your IV/CP settings due to moves being empty by default.

For example, in the debugger, I'm seeing that it will transfer the following pokemon, even I have ```"KeepMinCp": 1600``` set.

screen shot 2016-08-10 at 1 43 25 pm

actislav:
pls remove this PR - , merge my instead #4827, this have issue, read above.

PR #4787, opened 2016-08-10
Bug fix - Sniping from pokezz fixed after pokezz implementation change

jjskuld:
## Short Description:
Fixes pokezz sniping when pokezz website is down.   Currently, if pokezz is not responding, then the bot will just do nothing when it starts to snipe.  It just waits indefinitely.

Now I've added a 5 second timeout, so that bot continues along if pokezz is not responding within 5 seconds.

### DIFF
--- a/PoGo.NecroBot.Logic/Tasks/SnipePokemonTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/SnipePokemonTask.cs
@@ -554,7 +554,7 @@ namespace PoGo.NecroBot.Logic.Tasks
 
             var socket = IO.Socket("http://pokezz.com", options);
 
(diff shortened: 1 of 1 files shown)


jjskuld:
I just noticed that the reason Pokezz was causing the bot to hang was that they changed their implementation (no longer using JSON).  So I fixed the code so that Pokezz is working again.

slyfoxjosh:
Does this fix require that i download a newer version again or is it like a server side fix on your end?

MrPK2710:
what i do with it ?

MrPK2710:
Got it" Work know man! thx

Kozartu:
how to use this?

PR #4793, opened 2016-08-10
Fixes transfer pokemon whilst catching

cawk:
makes it transfer them after catch success


LordDeimos:
@cawk just a question, since it appear to make the transfer without using any delay after the catch. Intended or to make more human-like this should always wait the delay between actions?

PR #4827, opened 2016-08-10
Fixed transferring ALL pokemons if no moves in config

actislav:
All changes tested and has backward compatibility.
Fixed transferring ALL pokemons if no moves in config;
and merge resutt of multi move list filiter;
more accurate and clear calculation of keep pokemon count acording to evolve task;
add check for empty multi move list, on manner of usekeepminlevel;


actislav:
oh yes, i forgot to mention,  now move list will work if certain count of move intersects pokemon moves, like if you set only 1 move then move1 or move2 must be equal set value, if you set 2 move, then both must equal to move1 and move2, if you set 3 and more moves then any 2 of them must be equal

PR #4878, opened 2016-08-11
Add missing await calls

jjskuld:
Fixes up the compiler warnings:
```
Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call.
```


ChemiHatake:
This slows down the bot and that's why it was removed.

jjskuld:
So I am curious, what is the effect of removing those await calls?  Does that functionality work then?

ChemiHatake:
@jjskuld no, but it might lead to errors

PR #4954, opened 2016-08-11
[Bug Fix] TransferPokemonTask's do not RefreshCachedInventory before deciding what to transfer

mrnitrate:
## Short Description:
Currently TransferPokemons task's do not refresh the cached inventory before looking for pokemon to transfer.  
When EvolvePokemonTask finishes evolving it does not update the cache. 
This can cause TransferPokemon to attempt to transfer pokemon that do not exist anymore or not transfer newly evolved duplicate pokemon. 

All tasks to change pokemon and inventory should refresh cache before and after (or maybe just after?) their operations.

TransferPokemons already have Inventory.DeletePokemonFromInvById that updates the cache after a transfer but the other tasks do not have corresponding methods.

Then we dont have to RefreshCachedInventory in FarmPokemonTask (which is currently incorrectly checking whether or not to refresh, fortsearch is only the last fort not the previous 6-12).

### DIFF
--- a/PoGo.NecroBot.Logic/Tasks/EvolvePokemonTask.cs
+++ b/PoGo.NecroBot.Logic/Tasks/EvolvePokemonTask.cs
@@ -87,6 +87,7 @@ namespace PoGo.NecroBot.Logic.Tasks
                     }
                     await evolve(session, pokemonToEvolve);
                 }
+                await session.Inventory.RefreshCachedInventory();
             }
         }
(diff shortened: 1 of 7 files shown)


actislav:
I prefer to make cache invalidation just after tasks which are generating it inconsistency (because transfer as you wrote have own method to it without full refresh by request). 
Can you do it in that way?

PR #5037, opened 2016-08-11
Add Ability for Missed Throws

KOTR2003:
Been so many changes since I last touched this I need to start a new PR for it.


sooth:
Wonderful. I'm glad to see others are concerned about making the bot look as human as possible!

KOTR2003:
No doubt, exp per hour means nothing if the account just gets banned in a month

cawk:
am going to test this now and merge if it works

Issue #4862, opened 2016-08-11
settings use the term "pokemons", while the plural is also just "pokemon"


Marsinator:
## Stop! Before you create this issue (you can delete this section when opening the issue):
1. Have you validated that your config/auth.json and config/config.json is valid JSON? Use http://jsonlint.com/ to check.
2. Have you searched to see if there are other issues for the same issue? If so, comment on that issue instead.
3. If your question is referring to how to set up or use the bot, please join our discord instead of posting an issue: https://discord.gg/VXKxNFr

### Expected Behavior
settings using the term pokemon when refering to more than one

### Actual Behavior
settings refer to pokemons

### Your config.json
it really isnt needed, its the same for everyone.

one example: "PokemonSTransferFilter": 

### Steps to Reproduce
open config.json in editor

### Other Information
Version: 0.8.5
Windows Version: 10
What is affected (catch, transfer, evolve, recycle, snipe, powerup/levelup, other):
grammar

NozFiend:
Troll level = expert(s)

AdamWeishaupt1776:
I don't see how consistency in variable names is trolling. It improves code readability, maintainability, and overall sanity for everyone involved (devs and users).

Oolaruu:
Current version have those:

"EvolveKeptPokemonsAtStorageUsagePercentage"
"KeepPokemonsThatCanEvolve"
"PokemonsNotToTransfer"
"PokemonsToEvolve"
"PokemonsToLevelUp"
"PokemonsToIgnore"
"PokemonsTransferFilter"

zzzz9999:
I don't think pokemons is wrong.
In many languages adjectives plural ends with "s".
In En_en don't exist plural for adjectives. But exists for nouns.
So IMO Pokemons is ok.

One Dragonite
Two Dragonites
...
One car
Two cars
...

Oolaruu:
https://github.com/NECROBOTIO/NecroBot/pull/5136

Oolaruu:
Alright I submitted a PR who fix that issue.

CreativiTimothy:
I love Pokemon (plural) = I love boys/girls (plural)

Agree

NozFiend:
one fish, two fish
red fish, jew fish

Oolaruu:
http://lmgtfy.com/?q=pokemon+plural+form
Good reading.

PR #5136, opened 2016-08-11
Fixed for #4862 Pokemons to Pokemon settings name (config.json)

Oolaruu:
Fixed for #4862 
Pokemons to Pokemon for settings (config.json)
"EvolveKeptPokemonsAtStorageUsagePercentage":
"KeepPokemonsThatCanEvolve":
"PokemonsNotToTransfer":
"PokemonsToEvolve":
"PokemonsToLevelUp":
"PokemonsToIgnore":
"PokemonsTransferFilter"


jurannus:
This is a big nitpicky waste of time and effort and you clearly don't understand how our config system works - this change would break the settings for any user who had set custom lists in their config files. Will not be approved.

PR #5156, opened 2016-08-11
Fix pokewatchers sniping, which was not respecting the 1 req. every 5 seconds

jjskuld:
## Short Description:
Fixes Pokewatchers sniping not being reliable due to consistently getting error response from API:
```
You can only request the API every 5 seconds.
```
### The fixes:
1. Now it correctly retries up to 5 times, sleeping 1 second in between.  Note that pokewatchers sniping can now wait up to 5 seconds while it is checking the API.  After retrying 5 times, it gives up and moves on.
2. Also cleaned up the code to not modify the JSON response from pokewatchers.  The response modification was probably copied from the pokesnipers code.  I don't think it is necessary here (I did double-check that this was the case for Farfetched). I also checked code in the POGO Location Feeder project and did not see this modification.


jjskuld:
This is pokewatchers sniping fixed.

pokewatchers sniping

Without the fix, you will never see pokewatchers sniping happening, because almost always it is returning the error message, which is unparseable as JSON

jjskuld:
I'll attach my config and a running log shortly.  Will let it run for at least an hour before posting my log file.

PR #5189, opened 2016-08-11
Randomize next pokestop between first and second by distance 

FabryB:
## Short Description:
More human behaviour: not always choose the first pokestop in radius, but sometimes choose second one.

## Fixes:
This generate more random paths, because every time I restart the bot with same coordinates it would do the exact same path, which is not very human.


FabryB:
I see the label "PR Needs Testing & Verification", can I help providing something?

EG screenshot from Pokemon Bot Visualizer ?

GrimmGringo:
Yes please

ThomasSRL:
I would say doing the same path is human? It is what I would do if I ever bothered doing it legit. 
But sure more randomness is a good antiban for now.

Just dont add another option in the config please :D (just my opinion; far to many options as is)

PR #5224, opened 2016-08-11
[Snipe][Bug] Pokewatchers API cooldown

somani6:
Pokewatchers added a 5 sec cooldown to the api what causes HttpClient to
constantly fail.
Redone with WebRequest


jjskuld:
Looks like someone else is trying to fix this issue.  I've already coded it in #5156

@somani6  Ugh.. did you even test this?  There are several problems with your code:
1. You removed the special Cloudflare webclient that bypasses the DDOS protection.  Without that webclient, you will run into the Cloudflare protection screen when trying to access the API and you will get nothing back.
2. Looks like you just added some timeouts.. which does NOTHING to retry when you get the 5 second timeout error.

Don't blindly copy and paste code without understanding what it does.

PR #5240, opened 2016-08-11
Fixed next Pokestop ETA

somani6:
Calculation of the ETA to the next pokestop was wrong.


GrimmGringo:
That's not what it's doing ... at all.. the result is going to be WAAAY off.

somani6:
@GrimmGringo  why would it?
Distance is in Meters, Speed in km/h

For example 500m distance at 6km/h:
300 / 6 * 3.6 = 180sec => 3min witch is absolutely correct

PR #5326, opened 2016-08-12
Show pokemon list ar the time every 5 mins

eveliojosegarcialeon:
Show pokemon list ar the time every 5 mins...it's only shown when starting necrobot.


GrimmGringo:
You do realize there's no code here right

PR #5413, opened 2016-08-12
RENAME translation.ja-JP2.json>translation.jp.json

B100c:
This file does not occur DBCS bug.
translation.ja-JP.json in the Ver0.8.6 raises the DBCS display bug.
Please attach here by default in Japanese.


Oolaruu:
I do consider that we have to choose the correct japanese file who work, but the correct japanese language code is **ja-JP**

https://msdn.microsoft.com/en-us/library/ee825488(v=cs.20).aspx

B100c:
OK
Is it possible to replace the name of both of the file ?

Oolaruu:
I haven't check both files, but why not just have one ? the one who works ?

B100c:
translation.ja-JP good is here when translation is that has been removed in the future a bug because it output well to perfect a log file
translation.ja-JP2 is now hard to read in the current is a file in order to display correctly Japanese speaker but does not generate a bug
There is a need to reverse the name is translation.ja-JP2 should now default

Oolaruu:
If that translation.ja-JP2 is updated, display and run without a bug, then should be default.

B100c:
That's right
It must be written in single-byte character Japanese localization file now

PR #5436, opened 2016-08-12
fixed snipe bug of going to already expired pokemon location

shcheuk:
The bug is hided in a congested logic to determine if the locations from the web to be used to snipe. This fix separate it to easy to read logic and echo a message to user. ( user don't know which source (pokesniper or poke watcher or skipplagged ) is reliable and which source suit their need, so it is better let them know the actual filtering process.....For example, user use both pokesniper and pokewatcher may find the results from poke sniper is more than poke watcher but actually, it is because user called the pokesniper function first before pokewatcher, so most pokemon in pokewatcher is already visited during the pokesniper scan operation. With a status added to the filter process, user can know why they can't get any result from a certain source. 
Another example is, if the user wrongly config the IV level too high or opt not to hunt unknown IV pokemon accidentally. They have no clue why there is always no pokemon come from the web site.
Adding a status marking (already visited, IV too low, not in user snipe list, expired) making user easy to understand why there is no result or what is actually happening


ExcuseMi:
Can you add this for the PogoLocationFeeder server code as well?

PR #5508, opened 2016-08-12
Fix event exception

shcheuk:
## Short Description:
So many exception message floating in the diagnosis and output windows of visual studio....from the event and stataisticAggregator.....


shcheuk:
seems like accidetally mixed with another pull request........

crinq98:
so what should we do to solve the problem?

shcheuk:
Commit both of them will do.

crinq98:
http://go.microsoft.com/fwlink/?LinkId=780597 download this

PR #5559, opened 2016-08-26
Delays

nullum:
Suggesting a special delay for pokemon evolution, like 25 seconds (the time it takes in the game). 

And a delay variable for all the delays, so every delay is slightly different each time, not always 5 seconds per user action for example, but one time 4.97, another 5.38.... same for pokemon catch and co.

When the bot starts, would be nice to don't immediately transfer the Pokemon but with a random delay, same after "throttling api calls", because it looks too "botish".


SmilingSword:
i already request this long time ago. since like the dev deleted my section. hope it will be implemented soon :D

jaja228:
sadsaczsfveascvs

mtailordev:
FeelsBadMan

rafalohaki:
when new version?

